PR #180, opened 2026-02-18
Extract dag into a separate waymark-dag crate [1 / 4]

MOZGIII:
This PR utilizes the work started at #179 and extracts the dag into a separate crate.

To review changes in this PR only go here: https://github.com/piercefreeman/waymark/pull/180/changes

Now we can clearly see that dag has no dependencies on i/o (besides, maybe, std i/o) or tokio - thus it is easy to conclude it consists only of dag types and related conversions.

In the further work we could further split these two modules into `dag-core` (as in holds only the core types) and `dag-coverter` (some logic on top of core types); this kind of organization is easy to reason about.

Another potential future improvement is to split the ast (aka ir) `waymark-proto` into an `waymark-proto-ast` package.

This work also unlocks extracting backends base into a separate crate, since they depend on dag. It does not need the converter - only the dag types themselves - so this is where we'll see the benefit from `dag-core` / `dag-converter`  split in the future


piercefreeman:
@MOZGIII Might have to rebase for this one, otherwise looks good to me.

MOZGIII:
Ready to go

PR #183, opened 2026-02-21
Bump thiserror to v2 and unify it's versions [2 / 4]

MOZGIII:
Goes after #180


MOZGIII:
See https://github.com/piercefreeman/waymark/pull/195

PR #193, opened 2026-02-27
Refactor the code in peparation for extracting binaries as crates [9 / 4]

MOZGIII:
https://github.com/piercefreeman/waymark/pull/193/commits/d4ebd36abf99bc8d5db90d204187eacb98f90ffb

This PR moves cli implementations into actual bin targets (and starts using waymark as external dependency).
Also, we extract the observability setup crate from the observability mod for easy reuse across bin targets as separate crates and for better scoped feature-flags


MOZGIII:
See #201

PR #206, opened 2026-03-02
Switch to using a shared uv workspace for python code

MOZGIII:
This PR is stacked over #205.

This PR changes the code layout such that there is a single shared `.venv` for the whole repo (except for an example app).

The code we reorganize this way is:
- `python` dir
- ~~`scripts` dir~~ I see scripts want version `>=3.12` which interferes with python version variation testing; I'll undo the changes that include the scripts into the workspace

They now share a common `.venv` for the python version and dependencies - as it doesn't make sense to have the flexibility of supporting multiple python versions in the single worktree.

Note this doesn't affect how many python versions the `waymark` wheel itself supports, and we continue to test it with multiple versions at CI.

The example app is not unified in the manner because it there is some uncertainty if it intentionally needs the flexibility of custom python version - but it is more likely that it doesn't. @piercefreeman let me know if we should unify it as well.


piercefreeman:
> Great! What about the example app? Should we include it too?

Yeah - same principle should apply.

PR #214, opened 2026-03-04
Extract webapp into separate crates

MOZGIII:
Goes in after #215.

This PR extracts webapp from the waymark crate into a set of its own crates.

Here I've decided to apply certain additional refactors beyond just moving code around, since the webapp is a simpler and more stable part of the system.

Notably, I've started introducing the bringup crate, which replaces part of the webapp server.
The notion of a "server" is gone on the crates structure level, and instead we now have:
- routes (which are perfectly composable with other routes if we ever need this),
- config (which are types that are supposed to fully encapsulate the configuration specifics, and don't contain a lot of dependencies so they be used as part of crates that form configuration trees but aren't responsible for bringing the configured components up - typically config-file-structure-forming crates; but we use env vars so this isn't directly applicable),
- bringup (the code that depends on both config and routes and wires things through to expose a simple interface for launching the webapp; replaces `WebappServer`, but would hold it if it was still needed).

`WebappServer` type seemed to be more of a tradition-dictated thing than practical, so I removed it in favor of a simpler plain fn, and avoid introducing an "object"-like type without a justification.

This change loses the exposure of the effective socket address of the listen socket (it was not used anyway). It can be returned if we ever need it.

We now use cancellation tokens passed down from the outside (i.e. as part of the general cancellation context hierarchy), and the application termination code now properly waits for the web server to exit upon graceful termination, instead of just sending the shutdown signal and skipping the wait.

That last thing - how we spawn tasks and properly wait for all the spawned tasks to terminate - is another improvement that is very warranted here, but it is best to address once we'll be working on the bringup crates and bringup composition.

There is also a notable abstraction leak between the config and bringup code (and was effectively present even before this refactor) - I've left a TODO right in the code explaining the situation there.


crates/lib/webapp-routes/src/lib.rs
```diff
@@ -707,7 +619,14 @@ struct HttpError {
 
 impl IntoResponse for HttpError {
     fn into_response(self) -> Response {
-        let body = Json(serde_json::json!({ "message": self.message }));
+        #[derive(serde::Serialize)]
+        struct ErrorBody {
+            pub message: String,
+        }
```
piercefreeman: Nice call on typing this

PR #231, opened 2026-03-06
Reduce code duplication via waymark-ir-conversions

MOZGIII:
This PR consolidates the code that is related to the proto-to-json literal conversions at `waymark-ir-conversions`, and reuses it tree-wide instead of various reimplementations.

There is a bigger discussion to have with regard to:
- use of JSON and specifically `serde_json::Value` as an internal model (suboptimal?)
- use of protobuf-generated types as an internal model (also suboptimal?)
- there are way too many "foreign" types that represent/model the same concepts: we should consider rewriting the system without using protobuf-generated types & json, and only keep the conversion into those at the relevant transport boundaries; otherwise we can't have guarding new-types and, more generally, type-safe code; there are good ways to share the transport data-schemas that won't cause so much inconvenience at the code level
- introduction of the newtypes (as mentioned above) for primitives, and more heavy use of the type-system
- maybe we should make a full-blown grammar and lexer (see https://github.com/pest-parser/pest and https://github.com/rust-bakery/nom), and separate ast from the ir

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3753,6 +3753,7 @@ dependencies = [
  "waymark-core-backend",
  "waymark-dag",
  "waymark-dag-builder",
+ "waymark-ir-conversions",
  "waymark-ir-parser",
  "waymark-observability",
  "waymark-proto",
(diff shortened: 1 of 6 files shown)


piercefreeman:
@MOZGIII I suspect a full-blown grammar and lexer is overkill, mostly because we need to write custom compilers for each runtime client language anyway (and imo it's easier to think about typed objects there than it is to think about outputting another fully valid grammar). But let's file the rest for follow-ups.

PR #240, opened 2026-03-12
Runloop: refactor the available instance slots logic

MOZGIII:
Goes in after #236 

This PR continues the runloop refactors, now focusing on removing more logic from the `impl Runloop { ... }` blocks. We are introducing purposely-built primitive with a strict API for the tracking of the available instance slots.

### DIFF
--- a/crates/lib/runloop/src/available_instance_slots.rs
+++ b/crates/lib/runloop/src/available_instance_slots.rs
@@ -87,6 +87,7 @@ impl Tracker {
     /// Update the available slots by discounting the provided active instances
     /// value.
     /// Returns `None` if the number of active instances exceeds the capacity.
+    #[allow(dead_code)]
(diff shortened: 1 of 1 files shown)


crates/lib/runloop/src/available_instance_slots.rs
```diff
@@ -0,0 +1,71 @@
...
+    pub max_concurrent_instances: NonZeroUsize,
+}
+
+/// [`Tracker`] encapsulates an available instance slots calculation logic and
+/// a matrialized copy of computed available instance slots.
+///
+/// It also provies access to the available instance slots in an atomic and
+/// thread-safe way.
+pub struct Tracker {
+    pub calc: Calc,
```
piercefreeman: Is there some benefit to calc being public versus just used internally? Doesn't seem like the refactored code touches this at all (aside to instantiate it)... so I'm unclear where the responsibilities of Calc stops and Tracker starts.

PR #251, opened 2026-03-16
Adjust lints

MOZGIII:
This is small adjustment after #248


piercefreeman:
@MOZGIII I think let's just install tools before we run lint?

PR #289, opened 2026-03-26
Add waymark-jsonlines

MOZGIII:
This PR adds machinery for reading and writing serde-capable data items as JSONLines format.

I'm considering JSONLines format for replay feature logs - so the primary use-case is expected to be to stream action completions log.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4315,6 +4315,16 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


MOZGIII:
@piercefreeman I've reworked this crate, it is now much more idiomatic and less bloated

piercefreeman (approved):

PR #308, opened 2026-04-01
Bughunt branch [do not merge]

MOZGIII:
This PR is for the visibility of the new metrics and logs code we use for debugging production.

It may or may not be merged eventually, tbd.


MOZGIII:
This PR has been integrated into the main branch, with the exception of the sparse update graph that we tested but didn't find useful - he only commit left in this branch.

PR #342, opened 2026-04-13
Refactor hydration and how the dag is kept alongisde the instances

MOZGIII:
This PR addresses a few issues with how the DAG was managed at runtime.

I have made an observation that while the `dag` was an optional value of `QueuedInstance` it was never used anywhere but a small codepath that worked only with hydrated instances - not stored/loaded in/from the database, not used in the serialization - etc. This was a long-time annoyance, since before I did the audit of accesses to the dag for the purposes of doing this refactor it was unclear whether this was intentional or just a coincidence (although intuitively the contract was quite clear); for implementing the replay feature this was particularly important.

So to avoid any further doubts and to address this long-outstanding issue I've decided to short-list this refactor.

Here we split the `dag` value out of the `QueuedInstance` and add a new type `HydratedInstance` that wraps both `QueuedInstance` and the `dag` - which is not an `Option` anymore and is required to be present. This allowed for elimination of one error class that is now provably guaranteed to be impossible via the type-system (see https://github.com/piercefreeman/waymark/pull/342/changes#diff-c025325f46a7262c47bf26ef0590bad109f4608777c0c7f9f501b5647e5bd802L87-L88).

Notably, this eliminates the `waymark-core-backend` dependency on `waymark-dag` - which was unwanted and should improve the compilation graph.

I'll later consider if the runner state should have the same treatment - but for this for now is enough to reduce the space of possible things to handle / care about at VCR.

---

The essence of this PR is https://github.com/piercefreeman/waymark/pull/342/changes#diff-85a3d501d1ed7f6e67bb6b8455bd31764e6355e10b1b683a24bdad09ce614c68

### DIFF
--- a/crates/lib/runloop/src/shard/run.rs
+++ b/crates/lib/runloop/src/shard/run.rs
@@ -4,15 +4,12 @@ use tracing::{debug, warn};
 use waymark_ids::{ExecutionId, InstanceId};
 use waymark_worker_core::ActionCompletion;
 
-use crate::shard;
+use crate::{hydrated_instance::HydratedInstance, shard};
 
 #[derive(Debug, thiserror::Error)]
 pub enum AssignInstancesError {
     #[error("queued instance missing runner state")]
     QueuedInstanceMissingRunnerState,
-
-    #[error("queued instance missing workflow DAG")]
-    QueuedInstanceMissingWorkflowDag,
 
     #[error("start: {0}")]
     Start(#[source] super::executor::StartError),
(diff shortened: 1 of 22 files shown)


crates/lib/runloop/src/shard/run.rs
```diff
@@ -84,18 +83,6 @@ pub fn run_executor_shard(
                         continue;
                     };
 
-                    let Some(dag) = instance.dag else {
```
piercefreeman, line 87: nice

piercefreeman (approved):

PR #365, opened 2026-04-20
Introduce new core backend traits

MOZGIII:
This PR introduces the new core backend traits and typed metadata system.

The rationale for this change is that the current core backend has a lot of shortcomings:
- the errors are not properly and atomically typed
- the data types hold unnecessary values that are not needed for the core functionality (but can't be dropped cause they're needed elsewhere)
- only a big single trait it presented, which makes it hard to require concrete capabilities at various places at the code (i.e. an fn that only needs to be able to pull the queue doesn’t need to be able to add items to it - but we can't express that)
- the data types do not encapsulate the semantics for the operations exactly (i.e. we can use `Result` instead of two `Option`s)
- function APIs are designed in a way that may cause unnecessary clones

This PR proposes a new crate that solves all these problems. The integration plan is to gradually switch from the old system to the new one - by implementing it at relevant backends, and then integrating at the code as needed; once the switch is complete and the old core backend is dropped - we will also be able to optimize the database schema and improve this new approach even further.


crates/lib/core-backend-metadata/src/lib.rs
```diff
@@ -0,0 +1,90 @@
...
+
+#[typle(Tuple for 0..=5)]
+impl<Items> private::Sealed for Items
+where
+    Items: Tuple,
+    Items<_>: MetadataItem,
+{
+}
+
+#[typle(Tuple for 0..=5)]
```
piercefreeman, line 69: Can we add some more crate documentation about recommended uses of these new metadata objects?

piercefreeman (approved):
Looks like a good first step in this approach to me - and appreciate being able to lump the Either comparison logic in these core traits.

crates/lib/core-backend-metadata/src/lib.rs
MOZGIII: Sure; I've added some examples at the tests that kind of hint on how these arr supposed to be used. Will add some more explicit guidance though.

MOZGIII: Added docs to the metadata crate; will also document the new core backend too.

MOZGIII:
Added doc-comments to the core backend and made some adjustments based on the usage analysis of the current code using the original core backend; notably some fields were dropped entirely (`entry_node`), and some were replaced with another metadata field (action done timings).

PR #367, opened 2026-04-21
WIP Javascript workflow definition

piercefreeman:


MOZGIII:
Let's add this step-by-step, starting with a proper workspaces setup in the repo? Sharing my setup for reference: https://github.com/MOZGIII/template-web.

I think we'd want to have the `javascript/packages` be a workspaces directory, but with the `package.json` still existing at the repo root level.

PR #371, opened 2026-04-21
Add worker remote bridge service crate

MOZGIII:
Goes in after #370.

This PR adds the worker-bridge-service crate to fulfill the worker reservations and establish two-way communication to the connecting worker processes.

It operates similarly to how it is done at the `worker-remote` crate now, but is encapsulated and provides more explicit control flow management by the virtue of being implemented via `async-stream` vs doing `tokio::spawn`.

### DIFF
--- a/crates/lib/worker-remote-bridge-service/src/lib.rs
+++ b/crates/lib/worker-remote-bridge-service/src/lib.rs
@@ -1,0 +1,134 @@
+use std::{pin::Pin, sync::Arc};
+
+use async_stream::try_stream;
+use futures_core::Stream;
+use prost::Message;
+use tokio::sync::mpsc;
+use tonic::{Request, Response, Status, Streaming, async_trait};
+use tracing_futures::Instrument;
+
+use waymark_proto::messages as proto;
+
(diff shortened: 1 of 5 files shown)


crates/lib/worker-remote-bridge-service/src/lib.rs
```diff
@@ -0,0 +1,132 @@
...
+            Status::invalid_argument(format!("invalid WorkerHello payload: {err}"))
+        })?;
+
+        let worker_id = hello.worker_id;
+        tracing::info!(worker_id, "worker connected and sent hello");
+
+        // Create channels for bidirectional communication
+        // Buffer size of 64 provides reasonable backpressure while allowing
+        // some pipelining of requests
+        let (to_worker_tx, to_worker_rx) = mpsc::channel(64);
```
piercefreeman: nit: would be better as a global variable to make the magic constant a bit more centralized / obvious.

piercefreeman (approved):

crates/lib/worker-remote-bridge-service/src/lib.rs
MOZGIII: Added `MESSAGE_PROTOCOL_CHANNEL_SIZE` const

PR #372, opened 2026-04-21
Add worker message protocol crate

MOZGIII:
This PR is a part of the #278 series. Goes in after #368.

This PR pulls the low-level transport layer request/response communication protocol orchestration crate from the `waymark-worker-remote` crate. It is pulled mostly as-is, with the major refactors postponed until we finish merging the #278 chain.

In the future, we'll improve this part of the domain by unifying this piece of code with the parts that still remains at the worker side (see https://github.com/piercefreeman/waymark/pull/278/changes/45fdcc4adecdde5d79517a70646e6341f530b039).

We'll also split out roundtrip metrics calculations struct and apply composition instead of inlining the response data.

This is not done as part of this PR because we want to maintain a relatively similar code structure for this part of the refactor for now, and rather focus on the process/pool management and worker registration.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4925,6 +4925,19 @@ dependencies = [
 ]
 
 [[package]]
+name = "waymark-worker-message-protocol"
+version = "0.1.0"
+dependencies = [
+ "prost 0.12.6",
+ "thiserror",
+ "tokio",
+ "tracing",
+ "uuid",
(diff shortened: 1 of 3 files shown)


MOZGIII:
This is a redo of #369 cause I forgot to change the base branch to `main` before the merge, again.

PR #376, opened 2026-04-21
Add worker-remote-pool crate

MOZGIII:
Goes in after #375.

This PR adds the `worker-remote-pool` crate that wraps the `worker-process-pool` crate; this doesn't make much sense as it is tbh, and this implementation can, at the very least, just be inlined into the `worker-process-pool` - or reworked in even better and more modular way - but we're sticking with this approach for now to reduce the moving parts.

The general strategy at this point is to finalize the switch of the internals to the `waymark-worker-process` - and keep, for now, the logic in parts on top of it as-is.

Ultimately though, we should probably ave the following composable parts instead of this odd `worker-process-pool` / `worker-remote-pool` pair:
- `worker-pool-core` - a new abstract trait to capture the interface required for maintaining a pool of workers; the current `BaseWorkerPool` trait is something else, and should probably be renamed to something like `ActionExecutor` - where a possible implementation of it could be for a worker pool.
- `worker-process-pool` - a concrete implementation of the process pool that can maintain the workers and replace the missing ones, but with no extra fuss like roundtrip metrics or even sender maintenance
- `worker-pool-manager` - a generalized driver over any worker pool implementation that tracks and issues the worker recycling on `max_action_lifecycle`, allocates slots and keeps track of worker usage, and provides an implementation of that new `ActionExecutor` trait on top of any managed worker pool
- `measured-worker-pool` - a simple pool implementation that wraps another implementation and tracks and exposes the metrics for the worker pool - with granularity at the per-worker level
- `measured-action-executor` - a simple `ActionExecutor` implementation that wraps another implementation and tracks and exposes the metrics for the action executor (more high-level ones than the `measured-worker-pool` would be able to capture, but more portable, like would work in the VCR replay where we don't really have a pool of workers)

### DIFF
--- a/crates/lib/worker-remote-pool/src/lib.rs
+++ b/crates/lib/worker-remote-pool/src/lib.rs
@@ -1,0 +1,223 @@
+mod request;
+mod response;
+
+use std::{
+    sync::{
+        Arc, Mutex as StdMutex,
+        atomic::{AtomicBool, Ordering},
+    },
+    time::Duration,
+};
+
+use nonempty_collections::NEVec;
+
+use tokio::sync::{Mutex, mpsc};
+
+use waymark_runner_executor_core::UncheckedExecutionResult;
+use waymark_worker_core::{ActionCompletion, ActionRequest, WorkerPoolError, error_to_value};
+
+async fn execute_remote_request<Spec>(
+    pool: &Arc<waymark_worker_process_pool::Pool<Spec>>,
+    request: ActionRequest,
+) -> ActionCompletion
+where
+    Spec: waymark_worker_process_spec::Spec,
+    Spec: Send + Sync + 'static,
+{
+    let executor_id = request.executor_id;
+    let execution_id = request.execution_id;
+    let attempt_number = request.attempt_number;
+    let dispatch_token = request.dispatch_token;
+
+    let dispatch = match request::to_dispatch_payload(request) {
+        Ok(dispatch) => dispatch,
+        Err(short_circuit) => return short_circuit,
+    };
+
+    let before = std::time::Instant::now();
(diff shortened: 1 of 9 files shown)


crates/lib/worker-remote-pool/src/lib.rs
```diff
@@ -0,0 +1,219 @@
...
+        };
+
+        let Some(mut request_rx) = request_rx else {
+            return Ok(());
+        };
+
+        let pool = Arc::clone(&self.pool);
+        let completion_tx = self.completion_tx.clone();
+
+        tokio::spawn(async move {
```
piercefreeman, line 173: this spawn feels complicated enough to deserve a bit of comment-based narration of what we're doing

piercefreeman (approved):
Approved pending note

crates/lib/worker-remote-pool/src/lib.rs
MOZGIII: No problem, I'll add an explainer; FYI, it's directly migrated from the original implementation at `worker-remote`: 

https://github.com/piercefreeman/waymark/blob/4d7c6456f4b7d9a9e101ba23a8787c42387eda49/crates/lib/worker-remote/src/lib.rs#L1338-L1356

I figured I'd rather eliminate the need for a spawn like this entirely but I've made a decision to port these bits as-is right now, so I didn't bother touching it at all almost.

MOZGIII: Done, check it out: 
https://github.com/piercefreeman/waymark/pull/376/changes#diff-9443b72a6aa917bb7a959d5a40b9c0acf34d6de8631b2b3043e392fdc3648a2dR169-R172

PR #383, opened 2026-04-27
Increase grpc message limit

piercefreeman:
By default grpc has a 4MB limit, this PR expands that to 25MB to have a bit more headroom to handle large JSON payloads. 4MB is on the smaller side of legitimate action payloads for complex workflows. This PR is still a draft as we decide whether postgres can even handle TOAST sizes this large or if we're better off just requiring users to keep their payload more compact.

We could also consider gzip/brotli encoding if we want to minimize the payload size that's stored.

### DIFF
--- a/_typos.toml
+++ b/_typos.toml
@@ -8,6 +8,7 @@ extend-exclude = [
   # Generated files
   "python/src/waymark/proto/**",
   "target/**",
+  "**/node_modules/**",
   # Proto generated Rust code
   "crates/lib/proto/src/*.rs",
 ]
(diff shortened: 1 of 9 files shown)


_typos.toml
```diff
@@ -8,6 +8,7 @@ extend-exclude = [
   # Generated files
   "python/src/waymark/proto/**",
   "target/**",
+  "**/node_modules/**",
```
MOZGIII, line 11: Nit: should probably belong to a separate PR...

MOZGIII (approved):

MOZGIII:
I think this PR is a good idea regardless of whether or not TOAST can handle it simply because it provides the machinery to tweak the grpc message sizes and thus makes this factor explicit. We could set it to 4 mb but explicitly if we don't want 25 mb; these defaults are often easy to miss, so explicitly stating *something* is a value of its own.

MOZGIII:
Noting here for later, just in case: TOAST kicks in for values >2kb, so 4mb or 25mb - not much difference...

PR #384, opened 2026-04-27
Fail actions if grpc bridge is closed

piercefreeman:
In cases where the user's action return payload is larger than the grpc limit, the grpc internals will close the connection. Right now this just stalls out the action because it gets no ack of completion. We want this close to mark as a failure.


crates/lib/worker-message-protocol/src/lib.rs
```diff
@@ -32,6 +35,26 @@ struct SharedState {
         HashMap<u64, tokio::sync::oneshot::Sender<(proto::ActionResult, std::time::Instant)>>,
 }
 
+impl SharedState {
+    fn is_closed(&self) -> bool {
+        self.closed
+    }
```
MOZGIII: This fn seems redundant, accessing `self.closed` directly is meaningful enough; given this is a private type, it's best to minimize the footprint of the API

crates/lib/worker-message-protocol/src/lib.rs
```diff
@@ -220,6 +254,9 @@ impl Sender {
         // Register pending requests
         {
             let mut shared = self.shared.lock().await;
+            if shared.is_closed() {
+                return Err(SendActionError::ChannelClosed);
```
MOZGIII:
```suggestion
                return Err(SendActionError::SharedStateClosed);
```

Reusing the same error here would be super confusing for debugging

MOZGIII (approved):
Good find, good fix!

MOZGIII (changes requested):
I'm realizing now this would be a much better to put an `Option<SharedState>` behind the `Mutex` as opposed to doing this; the rationale is that we'll be modeling the same behavior in a completely type-safe manner without introducing a new API (we'd use a standard `Option` and an "open data-bag" struct).

crates/lib/worker-message-protocol/src/lib.rs
```diff
@@ -57,11 +57,28 @@ pub fn setup(channels: Channels) -> (Sender, impl Future<Output = ()>) {
...
+                }
+            };
+
+            if pending_acks > 0 || pending_responses > 0 {
+                tracing::warn!(
+                    pending_acks,
+                    pending_responses,
+                    "worker message protocol loop exited with pending requests"
+                );
+            }
```
MOZGIII: We should consider placing this under a `Drop`-guard inside of the loop; that way we'll also be closing the shared state on panics; we also probably can switch to an std mutex as there is no need to hold the mutex access awaits - that'll remove the issue of having a `.await` in the `Drop` impl

MOZGIII: I' thinking `struct SharedStateDropGuard(pub Arc<Mutex<Option<SharedState>>>);` and `impl Drop for SharedStateDropGuard { fn drop(&mut self) { let val = { self.0.lock().take() }; ... } }`

MOZGIII (approved):
Lgtm

PR #388, opened 2026-04-30
Initial version of the new VM

MOZGIII:
This PR adds the initial version if the new Virtual Machine for workflow execution.

The goal of this work is to eventually replace the current runner and DAG-based workflow executor. This is the first step in that direction, to be followed by a new compiler for workflows into this new VM's bytecode.

This design features:
- modular instructions sets support (we don't have a use case fore true modularity - so this is not that; it is, rather, just strong abstraction layers and isolation, which is very helpful for just having readable code)
- stackless coroutines and async/await support at VM/bytecode level
- register-based value storage
- explicit pull-based side-effects emission model
- tiny runtime API surface and simple and explicit run/step execution loop
- a simple API resolving the futures from the external code (~~currently requires mutable access, but maybe can be made non-mut-access based and thread-safe, we'll see~~ solved via vm-driver, fully wired up now)
- a cli for PoC demonstration and utilities (wip)
- no GC (yet), no support for referenced values (only by-move/by-copy passing can be supported without a GC of some sort)
- no type information at the VM/bytecode level (besides the variant of the Value enum)

More info on modularity:
- small runtime implementing only the abstract execution loop and pluggable per-instruction-set interpreters that hold the implementation logic for the executing instructions
- value type is abstracted, and specified at the instruction-set interpreter level (effectively allowing instruction-set interpreters to control the types of the values they want to use / support)
- type-safe trait contracts between the pieces, allowing us to express precisely what a certain part of the (rust) code has to know about the (rust) types it deals with
- code is organized such that the interpreter crates are separate from the instruction set crates, and, as an overall system part, interpreters could be dropped and replaced by a JIT compiler

---

Overall I'm pretty happy with how this implementation turned out; extending it with the full surface of the bytecode operations should be easy, and that's what I'm planning to do next, as it pretty much correlates with defining the full IR set, which is required to begin the work on the compiler.

---

Next steps and possible improvements:
- more instructions and allowed value types (scalars)
- GC and reference types
- resolved promise states cleanup (might require GC)
- `is_terminal(&self) -> bool` kind of API for the `Runtime` that would return `true` if there are no ready frames or prepared continuations; that'd cause some rewrite of the current promise states so it's posponed to later

Non-goals:
- stack unwinding and value bubbling at the VM level - we don't have stacks, and while we could technically add them it would inflate the state that we have to persist; rather than adding stacks, we can implement errors as explicit values and handle bubbling by injecting explicit bubbling instruction at the compiler level
- bytecode/VM-level type-checking - if we ever have to do the type-checking (as in for the program-defined type - as opposed to our own VM-level value-type enum, type checking for which we to either way then implementing the instructions) it has to be done at the compiler level
- instruction pointer at the frame level - we use states and state IDs for everything that that instructions are typically used for because we want to have only have explicit yield-points and continuations; this is not strictly a non-goal, but so far we've been able to avoid it, and it reduces the cognitive load as it's a competing duplicating concept

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -697,6 +697,40 @@ dependencies = [
 ]
 
 [[package]]
+name = "darling"
+version = "0.23.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "25ae13da2f202d56bd7f91c25fba009e7717a1e4a1cc98a76d844b65ae912e9d"
+dependencies = [
+ "darling_core",
+ "darling_macro",
+]
+
+[[package]]
+name = "darling_core"
+version = "0.23.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9865a50f7c335f53564bb694ef660825eb8610e0a53d3e11bf1b0d3df31e03b0"
+dependencies = [
+ "ident_case",
+ "proc-macro2",
+ "quote",
+ "strsim",
+ "syn",
+]
+
+[[package]]
+name = "darling_macro"
+version = "0.23.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ac3984ec7bd6cfa798e62b4a642426a5be0e68f9401cfc2a01e3fa9ea2fcdb8d"
+dependencies = [
+ "darling_core",
+ "quote",
+ "syn",
+]
+
+[[package]]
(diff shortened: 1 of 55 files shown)


piercefreeman (approved):
Was expecting this VM to be a lot more code, actually. The registry abstractions here are clean.

PR #392, opened 2026-05-06
Add an assert for type-safety at python/scripts/export_workflow_ir.py

MOZGIII:
This PR "fixes" a small python types mishap. Not sure why it doesn't work as-is, but the extra assert is redundant and shouldn't be necessary (since `parser.error` is typed to never return).

Is `ty` not smart enough to know that? Idk...


piercefreeman:
ty sometimes has differences in implementation to the mypy/pyright parsers (which are usually more comprehensive), but usually it's more conservative and accurate. In this case I think it's just a question of lacking sideeffect typehints for this function.

https://github.com/python/cpython/blob/970b0433f498fd0d048b1f50229a7f0fb4d9e07a/Lib/argparse.py#L2877

piercefreeman (approved):

PR #393, opened 2026-05-06
Initial compiler version

MOZGIII:
This PR adds an initial version of the compiler; this is a PoC-level implementation, has a very simple internal design that doesn't scale well in terms of features and code complexity and doesn't provide a lot of internal type-safety features or code reuse or testing infrastructure.

This is to be improved by #396 and #397. In fact, this whole work would normally be just squashed into this PR, but I've decided to provide #393, #396 and #397 separately to have some smaller and more readable diffs - and have snapshots in between the PoC and proper implementation, but without inflating this PR to introduce them.

This goes as follows:
- this PR (#393) adds the initial version of the compiler that can compile old python IR/AST to bytecode
- #396 adds snapshot tests to it, and contains some large diffs that aren't very convenient to read together with everything else
- #397 reworks the compiler by adding more structure and generally finalizing the work on it with docs and test coverage; it also has some small changes to the snapshots that show that the code changes don't change the codegen


crates/bin/vm-cli/src/sample/ast_old.rs
```diff
@@ -0,0 +1,27 @@
+pub fn program() -> waymark_vm_ast_old::Program {
```
piercefreeman, line 1: nit: would be helpful to just have some pseudocode for the functions in `sample` in the docstring

Not too hard to grep organically tho

piercefreeman (approved):
I've mostly reviewed the whole enchilada in 397 - but this is a solid stepping stone on our way there to teach the core concepts.

For future reference of anyone reading this:

- This keeps the waymark AST in its current form, assumes the Python->AST pipeline is maintained as is
- Compiles the AST into our new IR
- Lowers that IR into the virtual machine (frame based implemented in a previous PR)

crates/bin/vm-cli/src/sample/ast_old.rs
MOZGIII: Added the comments inline to keep them closer to the actual language items

PR #395, opened 2026-05-06
Preserve order of call definitions at the python AST

MOZGIII:
This makes the order of the call definitions emitted by the Python AST for workflows deterministic.

This could cause changes in the IR hash and, subsequently, lead to creation of the secondary version of a workflow for the same exact workflow file.

It was discovered while I was working on a new compiler and did the snapshot testing - they were flaky at the AST side - so here we are now.


piercefreeman (commented):
LGTM!

piercefreeman (approved):

PR #396, opened 2026-05-06
Add snapshot testing to the old-ast compiler

MOZGIII:
This PR adds snapshot tests to the compiler based on the tests from the python dir.

The code is best reviewed by hiding the `.snap` files first, and then reviewing them separately.

Look at `crates/lib/vm-compiler-for-ast-old/tests/python.rs` first.

That file holds the main test machinery and a `SKIP_TESTS` list - these are the tests files that our test invocation system doesn't yet understand; either they don't have a workflow or they have more than one - both of these cases are not properly handled just yet.

We have snapshots for both inputs and outputs for the compiler - so we can track the changes to specifically the compiler itself; if the tests files change we will have diffs for both the input and the output - explaining the change for the future recalls.


piercefreeman (approved):
Nice idea! Like this - and you're thinking of incorporating this check into our existing tests or to another CI phase?

MOZGIII:
It's already a part of the standard `cargo test` run - so just the same existing test flow; I guess we could extract it out if needed - but  I don't see a need for it

piercefreeman (approved):

PR #397, opened 2026-05-08
Refactor the compiler

MOZGIII:
This PR build on #393 and #396 and makes the compiler internal machinery structured and test-able.

This structure has proven to be quite efficient for representing the compilation state. Although it is not a proper multi-pass multi-step compiler just yet, it is already organized into small and reusable parts (plans and sub-compilers) that help avoid code duplication and provide separation of responsibility.


crates/lib/vm-compiler-for-ast-old/tests/python.rs
```diff
@@ -92,10 +93,8 @@ async fn compile_python_tests() {
         let details = format!("ast for python/tests/{}", python_test_file.display());
         insta::assert_debug_snapshot!(snapshot, program, &details);
 
-        let bytecode_result = waymark_vm_compiler_for_ast_old::compile::<
-            support::TestSpec,
-            support::TestLowering,
-        >(&program);
+        let bytecode_result =
```
piercefreeman, line 96: I'm seeing the bytecode snapshots at 54 Err vs 16 Ok, so the suite ends up mostly being a baseline of current failures. Fine for now but if we notice that there are some tests that fail in steady state, I would probably split expected-success fixtures from expected-unsupported fixtures.

piercefreeman (approved):
Clean refactor, especially with decomposed lowing.

crates/lib/vm-compiler-for-ast-old/tests/python.rs
MOZGIII: Yeah, it's a very rough wip for now; the idea was to get some minimal functionality working (of both the compiler itself and the snapshots). When we don't apply any manual grouping on top of them it actually becomes really clean workflow of fixing something and getting it flipped.
I thought eventually we'll just switch to doing an unwrap and only storing successes, as there aren't supposed to be failures at all with those tests I think... But yeah, once we get to that stage we could have a separate group that we unwrap_err instead and snapshot the errors.

While we're working actively on fixing them it's pretty much fix something, see the snapshots flip to `Ok` and see if things look right - so very manual inspection is intended there either way, and you explicitly commit the snapshot updates too.

---

Btw, there's not many features left to implement - I suspect I'll address at least a couple more things there tomorrow, move the amount of passing tests up.

PR #399, opened 2026-05-12
Implement support for binary and unary scalar ops

MOZGIII:
This PR goes in after #398 and add support for binary and unary operations to the VM.

### DIFF
--- a/crates/lib/vm-instructions-pureset/src/lib.rs
+++ b/crates/lib/vm-instructions-pureset/src/lib.rs
@@ -19,6 +19,29 @@ pub trait Spec: 'static {
     type ConstValue: core::fmt::Debug;
 }
 
+/// An unary operation.
+#[derive(Debug)]
+pub struct UnaryOp<RegisterId> {
+    /// The register to store the result of the operation at.
+    pub dst: RegisterId,
+
+    /// The register that contains the operand value.
+    pub src: RegisterId,
+}
+
+/// A binary operation.
+#[derive(Debug)]
+pub struct BinaryOp<RegisterId> {
+    /// The register to store the result of the operation at.
+    pub dst: RegisterId,
+
+    /// The register that contains the first operand.
+    pub a: RegisterId,
+
+    /// The register that contains the second operand.
+    pub b: RegisterId,
+}
+
 /// Pure instructions set.
 #[derive_where(Debug)]
 pub enum PureSet<Spec: self::Spec> {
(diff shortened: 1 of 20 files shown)


crates/lib/vm-interpreter-pureset/tests/integration.rs
```diff
@@ -35,13 +36,50 @@ impl From<TestConstValue> for TestValue {
...
+    fn add(
+        a: &Self,
+        b: &Self,
+    ) -> Result<Self, waymark_vm_interpreter_pureset::value::BinaryOperationError> {
         match (a, b) {
-            (Self::Int(a), Self::Int(b)) => Ok(Self::Int(a + b)),
-            _ => Err(waymark_vm_interpreter_pureset::value::AddError::NotAddable),
+            (Self::Int(a), Self::Int(b)) => Ok(Self::Int(*a + *b)),
+            _ => Err(
+                waymark_vm_interpreter_pureset::value::BinaryOperationError::UnsupportedOperation {
```
piercefreeman, line 56: Some of these absolute imports are becoming pretty long....

crates/lib/vm-instructions-pureset/src/lib.rs
```diff
@@ -54,6 +54,208 @@ pub enum PureSet<Spec: self::Spec> {
...
+
+        /// The register that contains the left operand.
+        a: Spec::RegisterId,
+
+        /// The register that contains the right operand.
+        b: Spec::RegisterId,
+    },
+
+    /// Apply Python-style logical `or` to two values.
+    Or {
```
piercefreeman: Is there a clean convention to have a centralized definition for dst/a/b for all of these binary operations? I can't think of one aside from a build time macro (a bit ugly)...

piercefreeman (approved):

crates/lib/vm-instructions-pureset/src/lib.rs
MOZGIII: We could do this:

```rust
struct BinaryOpInstruction {
        dst: Spec::RegisterId,
        a: Spec::RegisterId,
        b: Spec::RegisterId,
 }
 
 enum PureSet {
    Add(BinaryOpInstruction),
    Sub(BinaryOpInstruction),
    // ...
 }
```

crates/lib/vm-interpreter-pureset/tests/integration.rs
MOZGIII: Yep; it is very expressive; I was thinking about switching to using shorthands in more places; pushed to the max, that'd be something like `waymark_vm_interp_pset::val::BinOpError::UnsupportedOperation`... Still pretty long. I'm considering this before resorting to `use`s to just estimate how much we can gain from just adjusting the names and not changing the patterns around how we structure the code; but it's still long and substantially harder to read, imo.

I guess in this case the code is localized enough that we could harmlessly add a `use waymark_vm_interpreter_pureset::value::BinaryOperationError`. Should we change it?

crates/lib/vm-instructions-pureset/src/lib.rs
piercefreeman: Seems cleaner to me?

crates/lib/vm-interpreter-pureset/tests/integration.rs
piercefreeman: Wouldn't prioritize at the moment. Let's re-evaluate as part of a broader cleanup after this feature.

crates/lib/vm-instructions-pureset/src/lib.rs
MOZGIII: Yep, I'll change it that way; we should be able to make it fully transparent representation afaik when it comes to encoding/decoding these, so I think it's not going to be an issue.

crates/lib/vm-interpreter-pureset/tests/integration.rs
MOZGIII: Creating https://github.com/piercefreeman/waymark/issues/402 - let's add other notable things there to keep track

crates/lib/vm-instructions-pureset/src/lib.rs
MOZGIII: Check it out: [`76d5a75` (this PR)](https://github.com/piercefreeman/waymark/pull/399/commits/76d5a7500af730d86e4dc742c304318fa0ed824f)

The definition is more compact, the rest is a bit more verbose; also, we loose the ability to tweak individual instructions without "invasive" changes. Also it becomes non-trivial on how we'd place the code if we change the instructions... But the main question is will we ever expect to do something like that? I doubt it...

MOZGIII: Ended up with this design: [`385d105` (this PR)](https://github.com/piercefreeman/waymark/pull/399/commits/385d1051f946393255fc722dfe8820019a339497)

Here we kind of lean into the AST design and provide a compatible-ish implementation - and it simplifies things a lot code-wise. It's not a problem for us here, and it adds more structure - more than the original implementation in this PR, but also more than the in the comment above, and a more rigid and robust at that. this is a direct result of me considering whether we'll even need flexibility here and deciding it's most likely with worth it (even if we do new layered HIR/MIR/etc later - it won't cause issues, and extra structure will reduce the code further everywhere).

piercefreeman: This definitely makes more semantic sense to me, seeing as they're the same underlying numerical primitive.

piercefreeman (approved):

PR #401, opened 2026-05-12
Implement support for sleep at the VM

MOZGIII:
This PR adds a new `sleep` instruction in the extcalls set, and implements the compiler support for it.

The implementation utilizes our nonzero duration wrapper at the interpreter level.

 We don't support `sleep(0)` as an immediate "yield"s just yet. This is something we'd want to eventually support - but perhaps with a separate, explicit instruction instead of just sleep + await.

### DIFF
--- a/crates/bin/vm-cli/src/run.rs
+++ b/crates/bin/vm-cli/src/run.rs
@@ -74,6 +74,26 @@ pub async fn run(
                                 }
                             });
                         }
+                        waymark_vm_interpreter_extcallset::Effect::Sleep {
+                            promise_state_id,
+                            duration,
+                        } => {
+                            tracing::info!(?promise_state_id, ?duration, "sleep received");
+
+                            tokio::spawn({
+                                let promise_resolutions_tx = promise_resolutions_tx.clone();
+                                async move {
+                                    tokio::time::sleep(duration.get()).await;
+
+                                    let value = integration::SampleValue::Usize(0);
+                                    tracing::info!(?promise_state_id, ?value, "resolving sleep");
+                                    promise_resolutions_tx
+                                        .send((promise_state_id, value))
+                                        .await
+                                        .unwrap();
+                                }
+                            });
+                        }
                     },
                 }
             }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5318,6 +5318,7 @@ version = "0.1.0"
 dependencies = [
  "derive-where",
  "thiserror",
+ "waymark-nonzero-duration",
  "waymark-vm-instructions-extcallset",
  "waymark-vm-interpreter",
  "waymark-vm-runtime",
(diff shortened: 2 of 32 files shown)


crates/bin/vm-cli/src/run.rs
```diff
@@ -74,6 +74,26 @@ pub async fn run(
...
+                        waymark_vm_interpreter_extcallset::Effect::Sleep {
+                            promise_state_id,
+                            duration,
+                        } => {
+                            tracing::info!(?promise_state_id, ?duration, "sleep received");
+
+                            tokio::spawn({
+                                let promise_resolutions_tx = promise_resolutions_tx.clone();
+                                async move {
+                                    tokio::time::sleep(duration.get()).await;
```
piercefreeman, line 86: How are you envisioning this hooking up to the database layer for longer-term sleeps and self evicting from memory (ie. sleep 2 days)?

MOZGIII: I consider persistence, eviction and sleeps as three independent things.

First of all persistence:
- on each extcall (or a sequence of extcalls, if we need to improve the performance) we'd make a record into the database with the VM state dump;
- this will be implemented at the driver level, i.e. the driver will send over the state to persist so that the rest of the system doesn't concern itself when to actually perform the dump, or orchestrate it - it will just have to persist it;
- we can also make a state dump once we start an extcall
- the exact details are tbd, but this is the idea

Eviction:
- the VMs will linger for a fixed time period after they have last been executed, and once that period expires they'll be evicted
- there will be multiple "liveness locks" (terminology analogous to like a "wake locks" on mobile devices) that keep a VM instance from being evicted; so far I have two in mind:
  - lock that is active while we have unpersisted changes
  - lock that is active while less than X amount of time has passed since last time this VM was scheduled
- we'd want to avoid mixing terms "awake" and "alive"
	- "waking" is when we have a VM runner in memory that has no ready frames and we activate a continuation on it
	- "make live" (revive?) is when we have no VM runner instance in memory but we need one and we instantiate a new instance it from a persisted state dump
- the "sleep" will be nothing special, and will just naturally fit in this model
- to resume the execution from sleep, we'll have a special loop that resolves sleep promises when their time comes; it'd be a quite simple db polling mechanism
- we'll have a unified mechanism for "waking" the tasks when the promises they're subscribed to resolve; this will in short, consist of persisting the activation value in the DB (for crash resilience), making a VM runtime "live" (if we don't have an instance) and then resolving a promise into a running VM - and then it continues and executes on forward.

Sleep:
- already mentioned a dedicated loop above, but architecturally it's basically a special thing that serves extcall promises on the same level as workers
- but doesn't have to be external
- "sleep waker loop"

This will be a substantial change to how the runloop is organized, and will also, most likely require a couple more indirection layers / abstractions to be manageable and maintainable.

piercefreeman (approved):

PR #408, opened 2026-05-15
Implement length instruction / global function

MOZGIII:
This PR adds support for the `len(...)` global fn at the AST via a new `Length` instruction in the pureset.

---

The file sizes are getting out of control, so we'll do a pass on splitting them up in the near future.

### DIFF
--- a/crates/lib/vm-compiler-for-ast-old/src/function/compiler/bytecode/emitter.rs
+++ b/crates/lib/vm-compiler-for-ast-old/src/function/compiler/bytecode/emitter.rs
@@ -78,6 +78,11 @@ where
         self.emit(waymark_vm_instructions_pureset::PureSet::MakeDict { dst, entries }.into());
     }
 
+    /// Emits a container-length instruction.
+    pub fn emit_length(&mut self, dst: RegisterId, src: RegisterId) {
+        self.emit(waymark_vm_instructions_pureset::PureSet::Length { dst, src }.into());
+    }
+
     /// Emits a user-function call that writes a promise register.
     pub fn emit_call(
         &mut self,
--- a/crates/lib/vm-ast-old-helpers/src/lib.rs
+++ b/crates/lib/vm-ast-old-helpers/src/lib.rs
@@ -1,6 +1,7 @@
 use waymark_vm_ast_old::{
     ActionCall, BinaryOperator, Block, Call, ElifBranch, ElseBranch, Expr, FunctionCall,
-    FunctionDef, IfBranch, IoDecl, Kwarg, Literal, Program, Span, Spanned, Statement,
+    FunctionDef, GlobalFunction, IfBranch, IoDecl, Kwarg, Literal, Program, Span, Spanned,
+    Statement,
 };
 
 pub fn span() -> Span {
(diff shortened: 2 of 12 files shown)


crates/lib/vm-compiler-for-ast-old/src/function/compiler/bytecode/emitter.rs
```diff
@@ -78,6 +78,11 @@ where
         self.emit(waymark_vm_instructions_pureset::PureSet::MakeDict { dst, entries }.into());
     }
 
+    /// Emits a container-length instruction.
+    pub fn emit_length(&mut self, dst: RegisterId, src: RegisterId) {
```
piercefreeman, line 82: These global functions - like length - do feel a bit semantically different to me than our other VM pureset primitives.

Would a better encapsulation not be to explicitly introduce the idea of global vm functions (almost like internalcalls) and then just have the actual emitted pureset be a reference to the internal call?

MOZGIII: I've originally thought about this - i.e. builtin calls - but abandoned the idea for now.
Well, to be more specific - not abandoned the idea of the builtin calls pre se - but specifically for things like `len` and `append` I've decided to implement them as part of the core instructions set.

Adding an indirection layer and real global fns makes sense if we want them to be managed separately from the instruction sets, like make them user-provided for example. Then yeah, we'd do something like an `ActionCall` abstraction for them, but make them sync-only. I just didn't encounter any real need for that just yet though... Instructions are better because we can optimize the calling convention and also make them tailored to the specifics values.

Well, we could also have built-ins that are functionally equivalent to the instructions - but that'd be, pretty much, similar to grouping the global functions into an enum in the likes of `BinaryOps` - but with global fns it's not as justified since they don't all supposed to share arity.

We could move them to a separate instruction set - but that's also not quite fitting, since we might have only a handful of really low-level essential globals - I'm thinking just `len` and `append`; but maybe more...

I propose leaving things as-is for now; I suspect the best way of dealing with this is observing what emergent needs for grouping makes sense as we go.

piercefreeman: So your idea with implementing something like `enumerate()` in Python would be at compile-time, so we just create a for loop that tracks the index and then gets the object directly and assigns that to the vars?

piercefreeman (approved):

crates/lib/vm-compiler-for-ast-old/src/function/compiler/bytecode/emitter.rs
MOZGIII: I was just about to tackle for-loops and look into it; on the surface that was the idea, hence why I implemented length first; but it might not be possible to do with an index for all value types. We might need something else that's supported on a runtime level, but I'll need to investigate further.

We can actually implement it via lowering to a loop and call it a day - so that's the fallback plan that is known to work. That said - maybe it's not going to be too difficult to add a notion of a proper iterators or enumerators - at the very least for the VM-native values. But I want to do a bit of a research on how to make it right so the design fits not just python - but also, at the very least, EcmaScript and Ruby, and possibly Go.

Adding an instruction for `enumerate` as a global function is a possibility actually - for instance in cases where it's used in the assignment and not in a loop. Ultimately, I imagine we'd want to have the ability to have reasonable portions of language-specific standard libraries implemented in our VM; either have separate combinations of instruction sets, with some language-specific instruction sets as parts of the mix, or standard libraries as specific extension points and have a single instruction set with the ability to swap the language-specific stdlib reimplementations...

---

If we have built-in functions, they'd likely have to be implemented above the trait abstractions, such that they don't see real `Value` type and can't access it directly (i.e. only via interpreter value trait interfaces). That would be the common thing between them and instructions. But the question is whether they should be able to *introduce* new type bounds on the `Value`. Hmm, we should be able to do it actually, should, but it would be on another level, above (or below) the interpreters. The problem there is that there'd be no natural way for optimizing the lowering of them at the compiler side... Well, I mean, we could do that - but then again what would be the difference from an instruction. Also, unless we want to hard-code as much stuff in the proper compiler pipelines as we have it now: like, currently globals are a fixed enum rather than a dynamic-dispatch based call convention like action calls. If it remains like that - the natural way to support them at bytecode is also via a fixed enum as we have it now.

I guess global is not the right word for it btw - we really only care about the built-in language functions; whether they are global, or value-associated.

---

We also have to think about how to deal with the code that wants first-class functions support:

python
```python
bar = getattr(foo, 'bar')
bar()
```

javascript
```javascript
const foo = { bar: () => {...} };
const bar =  foo.bar;
bar()
```

ruby
```ruby
foo.send(:bar)
```

PR #409, opened 2026-05-15
Implement index and dict dot ops

MOZGIII:
This PR adds support for list index and dict dot ops.

Goes in after #408. But doesn't really use the length opcode machinery (turns out it's not necessary to wire things though it).

### DIFF
--- a/crates/lib/vm-instructions-pureset/src/lib.rs
+++ b/crates/lib/vm-instructions-pureset/src/lib.rs
@@ -162,6 +162,30 @@ pub enum PureSet<Spec: self::Spec> {
         src: Spec::RegisterId,
     },
 
+    /// Resolve an indexed access from an object and index value.
+    Index {
+        /// The register to store the result at.
+        dst: Spec::RegisterId,
+
+        /// The register containing the indexed object.
+        object: Spec::RegisterId,
+
+        /// The register containing the index value.
+        index: Spec::RegisterId,
+    },
+
+    /// Resolve an attribute access from an object and attribute name.
+    Dot {
+        /// The register to store the result at.
+        dst: Spec::RegisterId,
+
+        /// The register containing the accessed object.
+        object: Spec::RegisterId,
+
+        /// The attribute name to read from the object.
+        attribute: String,
+    },
+
     /// Build a list value from resolved registers.
     MakeList {
         /// The register to store the resulting list at.
--- a/crates/lib/vm-interpreter-pureset/src/value.rs
+++ b/crates/lib/vm-interpreter-pureset/src/value.rs
@@ -89,6 +89,34 @@ pub enum FromLengthError {
     ResultOutOfBounds,
 }
 
+/// An error from [`IndexOp::index`].
+#[derive(Debug, thiserror::Error)]
+pub enum IndexOperationError {
+    /// The value type does not support indexed access for the operands.
+    #[error("indexed access is not supported for these operands")]
+    UnsupportedOperation,
+
+    /// The provided index falls outside the bounds of the target object.
+    #[error("index is out of bounds")]
+    IndexOutOfBounds,
+
+    /// The target dictionary does not contain the requested key.
+    #[error("key is missing")]
+    MissingKey,
+}
+
+/// An error from [`DotOp::dot`].
+#[derive(Debug, thiserror::Error)]
+pub enum DotOperationError {
+    /// The value type does not support attribute access for this object.
+    #[error("attribute access is not supported for this value")]
+    UnsupportedOperation,
+
+    /// The target object does not contain the requested attribute.
+    #[error("attribute is missing")]
+    MissingAttribute,
+}
+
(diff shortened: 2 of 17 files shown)


crates/lib/vm-instructions-pureset/src/lib.rs
```diff
@@ -162,6 +162,30 @@ pub enum PureSet<Spec: self::Spec> {
...
+
+        /// The register containing the indexed object.
+        object: Spec::RegisterId,
+
+        /// The register containing the index value.
+        index: Spec::RegisterId,
+    },
+
+    /// Resolve an attribute access from an object and attribute name.
+    Dot {
```
piercefreeman, line 178: Is "Dot" the standard name for this operation in the VM world? To me dot is more of a syntax addressor like `my_dictionary [dot] my_key` that doesn't map exactly to my_dictionary[my_key]`

crates/lib/vm-interpreter-pureset/src/value.rs
```diff
@@ -89,6 +89,34 @@ pub enum FromLengthError {
...
+/// An error from [`IndexOp::index`].
+#[derive(Debug, thiserror::Error)]
+pub enum IndexOperationError {
+    /// The value type does not support indexed access for the operands.
+    #[error("indexed access is not supported for these operands")]
+    UnsupportedOperation,
+
+    /// The provided index falls outside the bounds of the target object.
+    #[error("index is out of bounds")]
+    IndexOutOfBounds,
```
piercefreeman, line 101: Reminding myself for the future that we'll want to adjust our python compilation for negative index values to do something like `len(items) - neg_index` ... unclear how we'll compile spans of indexes but I'm sure there's a way to do it via manually constructing a for loop over the indexes and writing to a new list.

piercefreeman: Just saw you have the negative case implemented here. Do you think for vm purposes we should disallow the negative syntax, and force that up to the compilation layer? @MOZGIII

MOZGIII: Not sure if we can catch it at compile time - we have no data whatsoever whether a variable value is positive or negative at compile time - i.e. if we want to reject the negative length... We could compile in a bytecode that does a check at runtime and passes a positive-only value though.

Doing it at runtime (on the VM side rather than via bytecode) is kind of trivial since we do the bound-checking anyway, and it is definitely less overhead. It is also more natural since we represent integers at signed `i64` values as opposed to unsigned values.

Actually, if we want to, we can have a different `Value` type that has a different set of rules - for instance that has explicit unsigned integers support. Different frontend languages will also have subtle differences in how operations work, so we can definitely do more than one implementation. One prominent example of this is that python modulo operation does not map cleanly neither to rust's `rem` nor to `rem_euclid` - which is kind of surprising; so, yeah; once we finish with the MVP of this new VM we'd have a bunch of things to think about. The architecture is pretty powerful now, so we can deal with most of the compatibility issues explicitly rather than weirdly hacking around them. Well, except, maybe, the compiler needs more work before we can reuse it properly...

MOZGIII: To address the actual question - I suggest leaving this at runtime for now; it'd just be quicker to move forward with this implementation, and I don't see any serious problem with keeping it around. We'd most likely compile the code into the bytecode in-flight, so we can be flexible about the changes to the compiler - meaning we chan change it later if needed.

piercefreeman (approved):

crates/lib/vm-instructions-pureset/src/lib.rs
MOZGIII: It's not the standard name, but I somehow thought it was petty well-known colloquially under this name; which seems to also be true but not at the interpreters/compilers.

Yeah, we could probably rewire dot notation though the same index instruction, and simplify the traits here. The only concern is things like a string: index access to a string can be different than attribute access. If the operand is a number then it's easy, but I think I recall is some language an index operator on a string with a string operand mean find a position of this substring in the string; most languages though that's a method call. So it could be different.

---

In python, we have this: https://docs.python.org/3/reference/datamodel.html#object.__getitem__

What would happen if we implement a `__getitem__` and do a `bar.foo` invocation? Is it equivalent to `bar["foo"]` and `bar.__getitem__("foo")`?

My knowledge of Python is insufficient here, but in Ruby for instance it's not equivalent: if you `def [](key)` it won't be called when a dot-access in invoked; to catch those you'd do `def method_missing(key)`.
In JS though it is equivalent.

MOZGIII:
I'll have think a bit more about representing the dot operation separately from the index operation internally... It is very natural to me to keep them separate, and it comes from the background in Ruby... But I need to internalize the context of our operation mode better and re-eval how my experience applies here.

MOZGIII:
I'm taking too long with this, so I'll postpone the decision to later; #412.

Merging this as is since the subsequent work needs to move further.

PR #410, opened 2026-05-16
Implement for loops support at compiler

MOZGIII:
This PR implements for loops support by lowering them into simple jumps at the compiler size.

The following modes are supported:
- iteration over lists (`for item in items`)
- iteration over ranges (`for index in range(...)`)
- iteration over enumerates over lists as positional (`for index, item in enumerate(items)`)
- iteration over enumerates over lists as kwarg (`for index, item in enumerate(items = items)`)
- iteration over enumerates over ranges as positional (`for index, item in enumerate(range(...))`)
- iteration over enumerates over ranges as kwarg (`for index, item in enumerate(items = range(...))`)

The last two don't make much sense at first glance, but if used with a stepped range could be useful for some cases.

Iterating over dicts is not supported currently.

For ranges, the supported variants are:
- `range(stop)`
- `range(start, stop)`
- `range(start, stop, step)`

Internally we support two modes of iteration:
- range-based
- index-based (with indexes traversing from 0 to the length of the list, exclusive)

---

This is what I had in mind for the initial MVP release of the compiler. I've looked into how we can add support for something more advanced, and it seems like it's going to be quite a bit of work. Not too much, but enough to postpone to later. especially since we don't seem to support anything more advanced just yet with the current DAG implementation.

---

With the for-loops out of the way, there is very little surface to cover left:
- spreads - looked into them deeper, and realized they're not as trivial as I thought; will probably work on them next
- exceptions - supporting exceptions as first-class values is easy, but providing bubbling and try-catch support can actually be a quite complex task, so I'll be doing some research in the background before tackling it; I mean, it could be done simply - but I wonder if there's a more elegant way
- minor corrections - like tuple unpacking in assignments and etc

---

Overall, the for-loops came to be quite clean; I imagined it would be more messy, but the encapsulate quite well.

### DIFF
--- a/crates/lib/vm-compiler-for-ast-old/src/function/compiler/lowering/for_loop.rs
+++ b/crates/lib/vm-compiler-for-ast-old/src/function/compiler/lowering/for_loop.rs
@@ -1,4 +1,32 @@
 //! For-loop lowering.
+//!
+//! `for` loops are lowered into a four-state bytecode skeleton modeled by
+//! [`ForLoopPlan`]: a `condition` state, a `body` state, a `continue` state
+//! (the target of `continue` statements and the natural fallthrough at the end
+//! of the body), and an `exit` state (the target of `break`). All variants
+//! share that shape and reach it through [`ForLoopCompiler::compile_loop_skeleton`],
+//! which lets each header shape contribute only the parts that actually
+//! differ:
+//!
+//! - **Header classification** ([`ResolvedForLoop`]) inspects the iterable
+//!   expression and decides between three lowering strategies: walking an
+//!   indexable value, a positive-step `range`, or a runtime-signed-step
+//!   `range`. `enumerate(...)` is unwrapped at this stage so the chosen
+//!   strategy is independent of whether the loop binds `(index, value)` pairs.
+//! - **Per-strategy lowering** ([`ForLoopCompiler::compile_indexed_loop`],
+//!   [`compile_positive_range_loop`](ForLoopCompiler::compile_positive_range_loop),
+//!   [`compile_stepped_range_loop`](ForLoopCompiler::compile_stepped_range_loop))
+//!   allocates the registers that must survive across iterations, then hands
+//!   the skeleton three closures that emit the loop condition, the
+//!   per-iteration body prep, and the continue-edge update.
+//! - **Variable binding** ([`compile_loop_bindings`](ForLoopCompiler::compile_loop_bindings))
+//!   normalizes the value/`enumerate` distinction and handles destructuring
+//!   into one, two, or N loop locals.
+//!
+//! Loop-control (`break`/`continue`) targets come from the shared
+//! [`LoopControlStack`]; the nested statement compiler used for the body
+//! pushes the current [`ForLoopPlan`]'s scope so inner statements jump to the
+//! correct states without knowing the lowering strategy.
 
 use waymark_vm_ast_old::{
     Block, Expr, FunctionCall, GlobalFunction, Kwarg, Literal, Span, Spanned,
@@ -242,7 +270,26 @@ where
         }
     }
 
-    /// Compiles a `for` loop that walks an indexable iterable.
+    /// Compiles a `for` loop that walks an arbitrary indexable iterable
+    /// (lists, tuples, strings, etc.) by stepping through `0..len(iterable)`.
+    ///
+    /// The lowering reserves three persistent registers up front:
+    ///
+    /// - `iterable_register` holds the evaluated source so we evaluate the
+    ///   iterable expression exactly once, even when the body mutates locals
+    ///   that the expression depends on.
+    /// - `index_register` is the loop counter, initialized to `0`. It doubles
+    ///   as the enumerate index when the binding is `Enumerate`, which is why
+    ///   no separate enumerate register is allocated and the continue update
+    ///   only increments `index_register`.
+    /// - `length_register` snapshots `len(iterable)` once via
+    ///   [`emit_length`](Self::context). Snapshotting matches Python's `for`
+    ///   semantics and avoids re-emitting the length probe per iteration.
+    ///
+    /// The body prep allocates a per-iteration temporary `item_register`,
+    /// emits `item = iterable[index]`, and routes both `item` and `index`
+    /// through [`compile_loop_bindings`](Self::compile_loop_bindings) so the
+    /// value/enumerate distinction is handled uniformly.
     fn compile_indexed_loop(
         &mut self,
         loop_vars: &[String],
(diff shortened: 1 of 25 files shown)


crates/lib/vm-compiler-for-ast-old/src/function/compiler/lowering/for_loop.rs
```diff
@@ -0,0 +1,765 @@
...
+        end: &'expr Spanned<Expr>,
+    },
+
+    /// `range(start, stop, step)`.
+    Stepped {
+        /// Starting value.
+        start: &'expr Spanned<Expr>,
+
+        /// Exclusive loop bound.
+        end: &'expr Spanned<Expr>,
```
piercefreeman, line 100: Nice, useful to have support for this too (believe the DAG actually doesn't at the moment).

crates/lib/vm-compiler-for-ast-old/src/function/compiler/lowering/for_loop.rs
```diff
@@ -0,0 +1,765 @@
...
+            } => self.compile_positive_range_loop(loop_vars, start, end, body, binding),
+            ResolvedForLoop::Range {
+                range: RangeLoop::Stepped { start, end, step },
+                binding,
+            } => self.compile_stepped_range_loop(loop_vars, start, end, step, body, binding),
+        }
+    }
+
+    /// Compiles a `for` loop that walks an indexable iterable.
+    fn compile_indexed_loop(
```
piercefreeman, line 283: This docstring (or inline ones) could benefit from a bit more of a description of what registers are involved here / how it will actually be laid out once compiled.

crates/lib/vm-compiler-for-ast-old/src/function/compiler/lowering/for_loop.rs
```diff
@@ -0,0 +1,765 @@
...
+        self.context.emitter.emit_jump(for_loop.condition_state());
+
+        let (exit_state, exit_flow) = for_loop.finish();
+        self.switch_to_with_flow(exit_state, exit_flow);
+
+        Ok(())
+    }
+
+    /// Compiles `range(stop)` and `range(start, stop)` loops with implicit
+    /// positive step `1`.
```
piercefreeman, line 332: Natural question is why not just call compile_stepped_range_loop with step=1. Can we justify that in the docstring? (my understanding is it's because of register allocation requirements)

piercefreeman (approved):
Agreed - clean implementation. Some of these compilation functions are pretty dense with register logic, to increase greppability could benefit from a bit more narrative flow either in the fn level docstrings or in intermediate comments. I'd opt probably for the docstrings to just give a high level sketch of the frame-pseudocode that we expect to pass to the VM.

crates/lib/vm-compiler-for-ast-old/src/function/compiler/lowering/for_loop.rs
MOZGIII: Added the justification:

https://github.com/piercefreeman/waymark/blob/4dbc0c199d8638e4088e7d67bc7d75cc27af4529/crates/lib/vm-compiler-for-ast-old/src/function/compiler/lowering/for_loop.rs#L294-L313

crates/lib/vm-compiler-for-ast-old/src/function/compiler/lowering/for_loop.rs
MOZGIII: I've tasked opus 4.7 to generate the docs, and it did a good job imo. Check it out.

My only concern is this is too verbose - I try to minimize the amount of explainers for two reasons:
- when the docs repeat what can be deduced in the code, their presence is redundant and only increases the amount of time needed to read the source; so I often try to actively reduce such redundancy by *removing* the docs
- with the above in mind, the reader can form a reasonable intuition that when the docs *are* present, it *is* worth reading them, as they won't have anything useless

But here, perhaps the docs we have right are useful.

I've refactored the code a bit to have more structure, and, hopefully, be easier to follow - so this might be enough to understand.

---

With a proper compiler I'd add something like sourcemaps to allow following what part of the original source the bytecode comes from; that'd help, among other things, how the compiler works in practice. For now my verification process consists of simply looking at the snapshots and "guessing" if the codegen makes sense.

crates/lib/vm-compiler-for-ast-old/src/function/compiler/lowering/for_loop.rs
```diff
@@ -1,4 +1,32 @@
...
+//! which lets each header shape contribute only the parts that actually
+//! differ:
+//!
+//! - **Header classification** ([`ResolvedForLoop`]) inspects the iterable
+//!   expression and decides between three lowering strategies: walking an
+//!   indexable value, a positive-step `range`, or a runtime-signed-step
+//!   `range`. `enumerate(...)` is unwrapped at this stage so the chosen
+//!   strategy is independent of whether the loop binds `(index, value)` pairs.
+//! - **Per-strategy lowering** ([`ForLoopCompiler::compile_indexed_loop`],
+//!   [`compile_positive_range_loop`](ForLoopCompiler::compile_positive_range_loop),
```
piercefreeman: Claude has a tendency to call out specific function names in its documentation, which I try to avoid because they're the most brittle thing that changes over time. If we prompt it to:

- remove these references
- instead give a narrative structure that provides /additional/ color commentary on how the flow of logic works and why we're doing it how we are
- and that it should specifically not just summarize the code logic that people can read

It usually does a pretty good job.

crates/lib/vm-compiler-for-ast-old/src/function/compiler/lowering/for_loop.rs
piercefreeman: This is okay, but I was thinking the ideal output would be lighter on prose. Ideally something that we could repurpose to other places in the compiler as the logic gets more complicated.

I'd think something like:

```

Raw code:

  results = []
  for item in items:
      processed = await process_value(item)
      results.append(processed)
  return results

VM registers:

  r0 = input items
  r1 = results
  r2 = iterable snapshot
  r3 = loop index
  r4 = length
  r5 = temp reused for cond/item/+1
  r6 = item local
  r7 = processed promise/value

VM compilation:

  S0: results=[], r2=r0, r3=0, r4=len(r2), jump S1
  S1: r5=(r3<r4), if r5 jump S2 else S4
  S2: r5=r2[r3], r6=r5, ActionCall -> resume S5
  S5: Await r7 -> resume S6
  S6: results += [r7], jump S3
  S3: r3=r3+1, jump S1
  S4: return r1
```

This is a bit brittle to the implementation but it provides us very easy readability of where the data is flowing, which is mostly what the compiler is responsible for here.

crates/lib/vm-compiler-for-ast-old/src/function/compiler/lowering/for_loop.rs
```diff
@@ -357,7 +404,31 @@ where
         )
     }
 
-    /// Compiles `range(start, end, step)` loops with runtime sign checks.
+    /// Compiles `range(start, end, step)` loops where the step direction is
```
piercefreeman, line 397: I like these prose ones more once we've established the general pattern.

crates/lib/vm-compiler-for-ast-old/src/function/compiler/lowering/for_loop.rs
MOZGIII: I figured this would make for sense to be in a test case than as the documentation for the actual code.

I also think we should have this, information available - but not in the form of a comment; rather, it would be better if we could emit some metadata about what registers represent what that is generated by the compiler itself. This seems to be the only missing bit of the information really. Also, this representation is really compact, and allows for much easier review of the code.

I don't like that we need to have to manually write this, and I'd rather work on adding the automation to cover our need with this here, than writing these cases out manually.

---

Since this test is already available I'll add it to the test cases suite now. But even then - setting the stage for manually keeping the comments like these in sync with the code seems very error-prone to me...

PR #411, opened 2026-05-21
Refactor the promises architecture at the VM runtime

MOZGIII:
This PR solves one big shortcoming that existed in the prior implementation of the VM: the promises were only available as the top-level register values; in other words, container values could not hold promises. With this PR, the promises are regularized within the value type, and the runtime no longer treats the separately at the type-system level. In short, `Promise<Value>` -> `Value` at the runtime level.

This approach allows us to put promises in containers; since we still need to have a notion of promises at certain code positions - like at the top-level return - we have a new system of traits that enables us to the corresponding trait bounds to expose this knowledge. This area is a bit of a WIP, as I'm still working on the right architectural seams between the contracts. A couple of places where we've inherited the limitations from the previous implementation still work as before, introducing unnecessary restrictions (mostly related to passing futures to the fns and etc).

This new design also shifts the responsibility of handling the promises to the interpreter value traits - which means that whether an operation is supported on a value or not will a value-dependent behavior that takes into account whether the value is a promise or not.

---

Unresolved questions:
- naming; we currently have the pattern where we have the following types:`Value`, `PromiseValue` and `ReadyValue`; the issue is with the `ReadyValue` - it seems quite awkward, as it, on its own, does not have to be aware of promises, but with the current composition it is, and is called `ReadyValue`; the alternative names considered are `ScalarValue` (doesn't fit cause it's not really scalars, containers too and essentially all the values) and `PlainValue` - which is also not right; ~~in fact, while writing this, I'm realizing this is actually a more advanced architectural problem than mere naming... anyhow; this is still a wip :]~~ nope, that went nowhere; I guess I'll leave it as-is for now

---

Future improvements:
- the value traits can have more places where we use associated error types, and the interpreter errors could wrap and expose those types in more places; for now we don't do it as there is not much need


piercefreeman (approved):
There are some nice properties with this refactor (that we might not be using at the moment but I imagine will be useful in the future):

- Containers no longer force accidental awaits, we can have a resolved wrapper type like Dict() or List() that itself holds promise values
- Pure operations can decide whether they require readiness; allows us to pass promises around to sub-functions (via copy) while disallowing unresolved promises while calling concrete functions (like len)

Did not dive into the tests as extensively as I looked at the business logic since it seemed refactored from the previous code... let me know if there's a section that would benefit from a closer look

MOZGIII:
The tests are just moving around to make them more manageable, as I found it quite inconvenient to navigate in the current state; also we're separating the harness/support code from the test fns.

MOZGIII:
> let me know if there's a section that would benefit from a closer look

@piercefreeman do you have any thoughts on the naming thing I've mentioned at the PR description? Besides that - not much I guess...

The `ReadyValue` name is the last thing I'd like to gather feedback before merging.

piercefreeman:
@MOZGIII I think it's okay. Maybe `ResolvedValue` gives more a sense that the high level container object has been resolved, even if there are some objects inside its containers that may be promises... but at least we know that the high level object is no longer a pending promise. ReadyValue doesn't scream technically incorrect to me either.

MOZGIII:
I've actually switched from `ResolvedValue` to `ReadyValue` as the latter seems more correct as a "standalone" term - i.e. the reasoning is what if the value is never wrapped in a promise - then it's not technically ever resolved, but just ready... I know it's maybe a weak argument but I made enough sense to me to make the switch.

I guess we'll go with this then. Maybe we should consider one other thing: renaming the `ReadyValue` to `Value` and using the `PromiseValue` everywhere; maybe with a `RootValue` type alias for the `PromiseValue`... But I'd experiment with that later.

PR #413, opened 2026-05-23
Add bytecode formatting facilities

MOZGIII:
This PR addresses some of the issues brought up during #410 review; this is not a 100% solution just yet, but it significantly offsets the difficulty of making sense of the snapshots during manual inspection.

The formatting is intentionally implemented to reuse the pre-existing facilities on a per-instruction level to avoid introducing more manual churn / code for supporting the formatting.

Suggestions on formatting are welcome!

---

In the future we may introduce a macro for deriving a custom / tweaked implementation for debugging the instructions; for now this does the job well enough though.

Another future improvement would be to have the AST code to pretty-formatted as well, but for that we'd need to implement a separate formatted as the existing one work with proto-generated types.

### DIFF
--- a/crates/lib/vm-bytecode-fmt/src/impl.rs
+++ b/crates/lib/vm-bytecode-fmt/src/impl.rs
@@ -3,15 +3,21 @@ use waymark_vm_bytecode_core::{FunctionId, InstructionId, StateId};
 
 use crate::Fmt;
 
+fn write_padding(f: &mut std::fmt::Formatter<'_>, level: usize) -> core::fmt::Result {
+    for _ in 0..level {
+        f.write_str("  ")?;
+    }
+    Ok(())
+}
+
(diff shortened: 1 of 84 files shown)


crates/lib/vm-bytecode-fmt/src/impl.rs
```diff
@@ -0,0 +1,78 @@
...
+use waymark_vm_bytecode_core::{FunctionId, InstructionId, StateId};
+
+use crate::Fmt;
+
+impl<'a, Instruction> Fmt<'a, TypedVec<InstructionId, Instruction>>
+where
+    Instruction: core::fmt::Debug,
+{
+    pub fn padded_fmt(&self, f: &mut std::fmt::Formatter<'_>, padding: usize) -> core::fmt::Result {
+        let pad = " ".repeat(padding);
```
MOZGIII: TODO: replace this with alloc-free fmt writer

piercefreeman (approved):
Nice! Way more readable

PR #417, opened 2026-05-24
Implement spreads support

MOZGIII:
This PR implements spreads support at the compiler. With #411 this is much easier to do (as it should've been really, it was a mistake to put the promises at the top-level of the registers).

Also, #413 and #414 really help with validation - it's much quicker to read the snapshots now.

---

With the VM, there is really no difference between spreading over actions and in-VM-fns; we might want to refactor the AST generation later to enable this capability.


piercefreeman (approved):
Most of the diff here is just in the snapshots + tests. The implementation of the spreads themselves is quite succinct; and will eventually be removed in favor of a cleaner spawn/await syntax but that's pending updates to the AST.

PR #423, opened 2026-05-29
Rework the Python worker process spec (#378)

### LINKED ISSUE #378: Rework how the Python worker process spec is implemented
Python worker spec has some issues - see the comments in the code of the `worker-python` crate.

It would be nice to make it more robust and also to rewrite it such that it does all the work to initialize the actual environment upfront (i.e. at boot time) and only captures the effective materialized values for the configuration when it's passed into the worker pool for operation. Doing this should also allow us to move the fallible and io-blocking initialization to the outside and eliminate panics (`unwrap`s) in favor of returning the errors and doing proper async-io as needed.

See https://github.com/piercefreeman/waymark/pull/375

intendednull:
Move all of the fallible I/O from config resolution to happen before worker spawn

Fixes #378


MOZGIII:
Thanks! I'll review this is a couple days, a bit busy currently.

MOZGIII (changes requested):
A but preliminary, as I didn't look too deep, but nonetheless some feedback. It seems like this PR also changes timeouts - I suggest we restrict this PR to python worker refactors only, as to avoid deliberation on the timeouts structure and etc; if possible of course.

intendednull:
yes that sounds good to me, thanks for the clarification

crates/lib/worker-python/src/config.rs
```diff
@@ -1,33 +1,22 @@
...
-    /// Arguments to pass before the worker-specific args
+    /// Arguments to pass before the worker-specific args.
     pub script_args: Vec<String>,
 
-    /// Python module(s) to preload (contains @action definitions)
+    /// Python module(s) to preload (contains @action definitions).
     pub user_modules: Vec<String>,
 
-    /// Additional paths to add to PYTHONPATH
+    /// Additional paths to add to PYTHONPATH.
```
intendednull: Unnecessary changes, avoid diff clutter please

crates/lib/worker-python/src/lib.rs
```diff
@@ -2,101 +2,109 @@
...
 pub use config::Config;
+pub use prepared::{PrepareError, PreparedConfig};
 
 /// Python worker process spec.
-// TODO: rewrite to fully cache effective values, like workdir, as constructor.
+///
+/// Holds a fully [`PreparedConfig`] plus the late-bound bridge address. All
+/// fallible/blocking setup already happened in [`Config::prepare`], so
+/// `prepare_spawn_params` is pure: it only assembles a command from cached
+/// values and never performs I/O or panics.
```
intendednull: Explanation leak, just document what this please. Link to the PreparedConfig docs already provides enough info

crates/lib/worker-python/src/lib.rs
```diff
@@ -2,101 +2,109 @@
...
-            // TODO: move to config
-            wait_for_playload_timeout: Duration::from_secs(15),
+            wait_for_playload_timeout: std::time::Duration::from_secs(15),
             shutdown_params: waymark_worker_process::ShutdownParams {
-                tasks_graceful_shutdown_timeout: Duration::from_secs(5),
-                process_graceful_shutdown_timeout: Duration::from_secs(5),
-                process_kill_timeout: Duration::from_secs(10),
+                tasks_graceful_shutdown_timeout: std::time::Duration::from_secs(5),
+                process_graceful_shutdown_timeout: std::time::Duration::from_secs(5),
+                process_kill_timeout: std::time::Duration::from_secs(10),
```
intendednull: reversion clobber here, make sure it matches original exactly

crates/lib/worker-python/src/lib.rs
```diff
@@ -2,101 +2,109 @@
...
+        tracing::info!(
+            ?reservation_id,
+            working_dir = %prepared.working_dir.display(),
+            python_path = %prepared.python_path,
+            "prepared python worker spawn params"
+        );
 
+        // Timeouts are left as the existing hardcoded defaults; reworking the
+        // worker-process timeout structure is intentionally out of scope here.
+        // TODO: move to config.
```
intendednull: Just leave the original comment, don't add more here

AGENTS.md
```diff
@@ -168,4 +168,6 @@ This section is used for the scratch updates, driven by our Agents.
...
 <rule>When shared async state must be closed during unwind or normal teardown, prefer a synchronous mutex plus a drop guard if the lock is never held across `.await`. Good: `let _guard = SharedStateDropGuard::new(Arc::clone(&shared));` with `Arc<std::sync::Mutex<Option<SharedState>>>`. Bad: async-only teardown paths that skip cleanup on panic.</rule>
+<rule>Don't run blocking filesystem/IO on the async runtime in a hot path; resolve and cache the materialized values once up front (e.g. an async `prepare()` that runs the probing under `spawn_blocking`) and have the hot path read the cache. Good: `Config::prepare().await` yields a `PreparedConfig` that a pure `prepare_spawn_params` reads. Bad: re-running `is_dir()`/`current_dir()`/`var("PATH")` on every spawn inside the trait impl.</rule>
+<rule>Don't `unwrap`/`expect`/panic on an infallible trait method or hot path; push the fallible/blocking initialization into a separate async constructor that returns a typed error. Good: `std::env::current_dir().map_err(PrepareError::CurrentDir)?` inside `prepare()`. Bad: `std::env::current_dir().expect("...")` inside `prepare_spawn_params`.</rule>
```
intendednull: wording on this might a bit confusing. more like "prefer fallible paths in failure-tolerance scopes, before they are passed into hot or infallible paths"

intendednull (commented):
@claude looks like the config revert clobbered a few lines. lets make sure the reversion is clean, no trace left behind

crates/lib/worker-python/src/config.rs
intendednull: Reverted the trailing-period doc tweaks on those three unchanged fields — they now match main exactly. (a476aa0)

crates/lib/worker-python/src/lib.rs
intendednull: Trimmed to a single line; the `PreparedConfig` link carries the rest. (a476aa0)

crates/lib/worker-python/src/lib.rs
intendednull: Restored the timeout block to match main byte-for-byte (`Duration::from_secs`, re-imported `Duration` instead of fully qualifying). (a476aa0)

crates/lib/worker-python/src/lib.rs
intendednull: Dropped the added explanation — back to just the original `// TODO: move to config`. (a476aa0)

AGENTS.md
intendednull: Reworded to your phrasing. (a476aa0)

crates/lib/worker-python/src/config.rs
```diff
@@ -70,29 +70,58 @@ fn default_runner() -> (PathBuf, Vec<String>) {
     )
 }
```
intendednull: Let's define this in the same place it is used. This file is more of a structural definition now, while the preparation work is done elsewhere, and this is part of the preparation work.

crates/lib/worker-python/src/config.rs
intendednull, line 76: Actually all of this can be moved now since there is a better home for it

intendednull (commented):
A small organizational nit. Please mark resolved when done

crates/lib/worker-python/src/spec_factory.rs
intendednull, line 1: This file is an extension of the existing config definition. Lets make it a sub-module of config instead of its own module. Use folder structure following existing project patterns

piercefreeman:
@intendednull Thanks for the PR! Waymark's still a really new project and we haven't yet formalized our contributor or AI guidelines yet. Will circle back to this ticket once we get that nailed.

crates/lib/worker-python/src/lib.rs
```diff
@@ -2,91 +2,60 @@
...
-        }
+        command.env("PYTHONPATH", &factory.python_path);
+        command.current_dir(&factory.working_dir);
+
+        tracing::info!(
+            ?reservation_id,
+            working_dir = %factory.working_dir.display(),
+            python_path = %factory.python_path,
+            "prepared python worker spawn params"
+        );
```
intendednull: Here we are doing SpecFactory -> Command. Lets factor this out. It could either be a standalone fn, or maybe it would make more sense to attach to SpecFactory. Lets follow existing patterns here, whatever practice is favored in this codebase. I'm leaning towards standalone, so lets just do that for now if unclear.

crates/lib/worker-python/src/spec_factory.rs
intendednull, line 1: I think this file name is too ambiguous. Lets do "spec_factory.rs"

crates/lib/worker-python/src/resolve.rs
```diff
@@ -0,0 +1,262 @@
...
+    // state to corrupt, so a clean error is preferable to aborting.
+    // (Contrast worker-process, which `resume_unwind`s associated-task panics.)
+    tokio::task::spawn_blocking(move || probe(config))
+        .await
+        .map_err(ResolveError::Join)?
+}
+
+/// The blocking half of [`resolve`]. Must only be called from a blocking
+/// context (e.g. inside `spawn_blocking`).
+fn probe(config: Config) -> Result<SpecFactory, ResolveError> {
```
intendednull:
```suggestion
fn resolve_blocking(config: Config) -> Result<SpecFactory, ResolveError> {
```

intendednull (commented):
@claude minor adjustments, please fix

intendednull:
If desired I'd be happy to spin up a new branch with a cleaner history. Not sure project preference here

PR #422, opened 2026-05-26
Add exceptions to the VM (via compiler-generated checks)

MOZGIII:
This is an entirely compiler-side implementation of the try/catch support, with exceptions being represented as a kind of value and bubbling implemented by checking whether the value is an exception and short-circuiting the function and returning the value to the caller.

This PR is superseded by #424.

PR #424, opened 2026-06-03
Add support for exceptions at the VM runtime and compiler

MOZGIII:
This PR supersedes #422 with a different (and arguably better) design.

---

Implements exceptions support at the VM runtime and compiler.

The exception is handled via a special path, as opposed to being passed around as a value; while it is possible to represent the exception as a value, it turns out we still need a distinction between an exception being passed around (i.e. returned from a function call) and the exception being bubbled (i.e. when the exception has been raised).
The #422 is the variant of the exception support that is much simpler, but lacks this distinction, and thus unable to represent the control flows we can encounter in the wild and would want to support.

In this implementation, `Exception` is separate type (that can be upcasted to a `Value` and downcasted from it in the contexts where we deal with values). This is similar to #422. What's different is we also have a `Raise` instruction, and an explicit notion of exception handlers, and thus we don't have to implement `IsException` global (although we could add it). #422 has an implementation that we can borrow.

---

#422 Implements bubbling by compiling in a check at every potentially-exception-generating instruction whether the value returned is an exception or not, and if it is an exception, it short-circuits by returning the value to the caller. The caller has the same check compiled in, and that's how the errors bubble.

The downside of #422 is:
- we can't represent an caught exception value being passed around as a regular value (we can actually - if we add a flag to it of whether it should be treated as "raised" or a value)
- there are a lot of compiler-in checks that don't really make sense; this complicates the codegen
- the API for the `Complete` effect - the separate `UnhandledException` effect it better

The upside of #422 is:
- the code at #422 is much simpler at the VM side
- only the compiler has to know where exceptions can come from, and it can, in theory, generate less checks; on the contrary, in this implementation, the knowledge of the compiler 
- separate instruction set (`excset`) - but it's not an architectural issue, we can do the same with this design, it's just cleaner in the coreset when we don't need as much extra instructions
- that implementation is, overall, built on simpler primitives, and relies on building things from smaller building blocks as opposed to this one, which hardcodes exception handlers

---

Now, this implementation has adds a single slot for a "raised exception" to the frame itself, and extends the runtime with the ability to execute interpreter-defined hook when we enter a state.
We implement the bubbling with those hooks - we check the exception handers first, and, if they're absent, we directly complete the whole runtime execution by emitting an `UnhandledException` effect (same semantics as `Complete`).
Any interpreter can inject a hook, but I've decided to implement the exception handling in the coreset just to move quicker for now; we can later on refactor it into a separate set if we have to. The architecture allows us to separate the two quite nicely.

---

#422 also has a different API for resolving/rejecting the promises (result vs separate fns), and a `Result`-based `Complete` effect vs separate `UnhandledException` effect. Test changes are nice and we should eventually bring them in from there as well; resolving/rejecting API is better here. `UnhandledException` effect is good too.

---

This finally completes the full scope of the implementation of the VM and compiler for the old AST - besides some missing built-ins (aka global fns) - but we might not really need them.
Next we can work on integration.


crates/lib/vm-compiler-for-ast-old/tests/snapshots/python__fixtures_control_flow__try_except_capture.py__bytecode.snap
```diff
@@ -1,11 +1,28 @@
...
+fn main(input: [], output: []):
+    try:
+        result = @try_except_capture.risky_action_capture()
+    except ValueError as err:
+        result = @try_except_capture.handle_error_capture(error=err)
+    return result
+---
+f0: [2 registers]
+  s0:
+    CoreSet(PushExceptionHandlers { handlers: [ExceptionHandler { handler_state: s2, exception_types: ["ValueError"], exception_dst: Some(r0) }] })
```
piercefreeman, line 14: New snapshot syntax is super helpful for these kinds of reviews.

crates/lib/vm-compiler-for-ast-old/tests/snapshots/python__fixtures_try_except__try_with_action.py__bytecode.snap
```diff
@@ -1,11 +1,25 @@
...
-    Unsupported(
-        Statement {
-            kind: TryExcept,
-        },
-    ),
-)
+fn main(input: [], output: []):
+    try:
+        result = @try_with_action.try_action()
+    except ValueError:
```
piercefreeman, line 8: Maybe I'm missing it, but it looks like we don't have any snapshots that cover exceptions as values? Can we add one?

crates/lib/vm-compiler-for-ast-old/src/function/compiler/lowering/statement.rs
```diff
@@ -138,6 +155,95 @@ where
...
+
+                match &mut continuation_flows {
+                    Some(flows) => flows.push(flow_state),
+                    None => continuation_flows = Some(NEVec::new(flow_state)),
+                }
+            }
+        }
+
+        if let Some(continuation_flows) = continuation_flows {
+            self.switch_to_with_flow(join_state, FlowState::union_branches(continuation_flows));
```
piercefreeman, line 241: Doesn't union here mark `value` as implicitly initialized; would this be a problem downstream?

```python
try:
    value = await risky()
except ValueError:
    pass
return value
```

crates/lib/vm-compiler-for-ast-old/src/function/compiler/lowering/statement.rs
```diff
@@ -138,6 +155,95 @@ where
...
+                    .register()
+            });
+
+            try_handlers.push(waymark_vm_exception_handler::ExceptionHandler {
+                handler_state,
+                exception_types,
+                exception_dst,
+            });
+        }
+        let try_exception_handler_depth = self.exception_handler_depth + 1;
```
piercefreeman, line 191: Instead of manually managing this unwind depth level, would it be cleaner to encapsulate into  a wrapper?

```rust
#[derive(Clone, Copy)]
struct UnwindTarget {
    state: StateId,
    exception_handler_depth: usize,
}

impl UnwindTarget {
    fn new(state: StateId, exception_handler_depth: usize) -> Self {
        Self {
            state,
            exception_handler_depth,
        }
    }
}

struct LoopScope {
    break_target: UnwindTarget,
    continue_target: UnwindTarget,
}
```

crates/lib/vm-compiler-for-ast-old/src/function/compiler/lowering/statement.rs
MOZGIII: Should be fine actually. This check is just to refuse the program at compile time. In Python, it should work the same - the interpreter will accept the program, but if `risky` fails with `ValueError` you'd get an `UnboundLocalError`.

https://www.online-python.com/rXRVz4tJc6

<details><summary>Details</summary>
<p>

```py
import asyncio

async def risky():
    raise ValueError()

async def f():
    try:
        value = await risky()
    except ValueError:
        pass
    return value
    
asyncio.run(f())
```

</p>
</details> 

In our VM you'd get a similar error - register unset... Unless the `value` is bound to an already initialized register - but *that shouldn't happen in practice right now*... Which is not good enough.

---

But maybe it is good enough for now. We have a persistence work next, which might force us to work on register pruning; that, in turn, might involve more advanced initialization and deinitialization mental model at runtime. We'd need a `Deinit` instruction for this as a prerequisite.

The #422 has a somewhat better compiler-side understanding of the handlers here, but in this variant I deliberately cut it to just the numeric depth. I'll do another look - cause maybe there are issues with more complicated scenarios in this variant (i.e. exception handler nesting and shadowing).

Also, the compiler could actually simply insert an explicit exception raise into the `pass` branch in theory... But that capability would need some new instructions - cause bytecode can't make new exceptions atm.

---


Btw, the old AST doesn't have a `raise` keyword - but I had to add it to complete my mental model at the VM level; there's no compiler-side support for `raise`.

MOZGIII: Or - we could reject this program at compile-time; cause that's what we had before I fixed it: [`171950a` (this PR)](https://github.com/piercefreeman/waymark/pull/424/changes/171950a495e5074e1e6a8d0d7263b8d321cfd473)

crates/lib/vm-compiler-for-ast-old/tests/snapshots/python__fixtures_try_except__try_with_action.py__bytecode.snap
MOZGIII: You are right; I'll add some more cases...

MOZGIII: [`4edbe83` (this PR)](https://github.com/piercefreeman/waymark/pull/424/changes/4edbe839a7bb7898cc43471b1080bf65f623aadc)

piercefreeman (approved):

crates/lib/vm-compiler-for-ast-old/src/function/compiler/lowering/statement.rs
MOZGIII: While it does help to follow the depth for unwinding, it also somewhat reduces the expressiveness. I.e. currently the code captures what's going on structurally better than if we couple depth with the target. Semantically, we also have a single variation there as opposed to two distinct possible cases (which would be captured by the proposed variation better). So, I'd leave it as-is.

MOZGIII:
I've addressed the last thing (https://github.com/piercefreeman/waymark/pull/424#discussion_r3367390257) and am planning to merge this in later today unless other things appear.

Issue #402, opened 2026-05-13
Notes for the broad clean-up


MOZGIII:
- https://github.com/piercefreeman/waymark/pull/399#discussion_r3235053221

MOZGIII:
Regarding the naming: noting here for the future discussion that I'd maybe swap the `ast-old` name with something else that is more specific; the compiler implementation had to be qualified by design, as we're supposed to have other implementation(s) and, possibly, universal implementation parts that are unqualified.

MOZGIII:
The generic bounds on the runtime are a bit too verbose; there are two things to do there:
- bound them all onto a single generic with associated types - we will do that if the second thing isn't effective enough
- actually reduce the amount of separate named bounds by grouping them more atomically (we should do it, but it's more of a straightening up of the internal contracts between things rather than explicitly addressing verbosity; shouldn't be abused and shouldn't be bet on to get big results from)

PR #437, opened 2026-06-23
Advanced VM driver facilities

MOZGIII:
This PR adds advanced facilities for properly driving forward new VMs.

- `vm-driver` contains the thread-safe shared async execution loop - the main loop that drives the runtime forward and allows the outside world to communicate with it; encapsulates and compactly expresses the persistence guarantees we aim to have in our design;

- `vm-driver-core` defines interfaces through which the outside world is supposed to communicate with the VM runtimes driven by the `vm-driver`;

- `blocking-future` crate is a low-level facility for executing "blocking futures" safely; this is intended for the futures - i.e. async functions that require access to an async runtime - but that, at the same time, can block: for instance on a blocking i/o or a tight CPU loop; the latter is a possibility when running arbitrary user-provided executables in our VM, and this is what justifies the existence of this implementation;

- `vm-driver-thread` crate wraps `vm-driver` loop with `blocking-future` and provides a thin glue code layer for regularizing the errors and restricting some unwanted access to the raw thread control APIs;

In the end, this PR improves upon the pre-existing control loop in the `vm-driver`, eliminates the async reactor thread pool blocking concerns, and integrates all this into the `vm-cli` sample executor.

---

There is a scary amount of generic trait bounds - this in itself is actually *fine*, since it's going to be simpler in other places, and is actually required to be this explicit for this particular location; the possible improvements in this domain are tracked at https://github.com/piercefreeman/waymark/issues/402#issuecomment-4639796320
In the long-run we'll improve on this substantially later, as we stabilize the trait sets. The architecture is still evolving here, and having explicit and precise control over the underlying details pays off.

---

Notably, this PR also updates the API at the runtime, exposing access to the internals like `waiting_promise_state_ids`.

---

This new design of the `vm-driver` essentially replaces the major parts of the runloop. Persistence safety is guaranteed via the `Ack` on promise resolutions - the happens before semantics would block the pending action result cleanups before we persist a snapshot of the VM runtime state that internalizes them.

The concrete mechanism for action persistence is not provided - but the interface it would have to fit is pretty straightforward and would be easy to implement with a buffering demultiplexer for database writes.


crates/lib/vm-driver-thread/src/lib.rs
```diff
@@ -0,0 +1,220 @@
...
+        ExecutionError,
+        SnapshotSerializationError,
+        SnapshotPersistenceError,
+        EffectHandlingError,
+        GettingPromiseSettlementsError,
+    >
+{
+    /// Abort the driver thread, cancelling the task.
+    pub fn abort(&self) {
+        self.task.abort();
```
piercefreeman: iirc there are some oddities with tokio and thread cancellations. Can blocking tasks be aborted once the blocking closure has started? Is this intended to be used only before the thread has started?

crates/lib/vm-driver/src/lib.rs
```diff
@@ -91,38 +126,110 @@ where
...
+        // Then, execute all ready frames until none are left and we suspend
+        // or there is an effect.
         match runtime.run() {
             Ok(effect) => {
                 tracing::info!(?effect, "effect");
-                if effects_tx.send(effect).await.is_err() {
-                    return Err(Error::EffectSenderClosed);
-                }
+                effector
+                    .handle_effect(effect)
```
piercefreeman: We seem to be pairing the `.run()` - ie. the generator of new frames - with the `.handle_effect()` - ie. the function that touches the outside world and kicks off the extcalls.

Let's say we kick off all these effects and crash before we get to the next iteration of the loop. So actions have been kicked off but we haven't yet persisted the frames that drive them. What's our recovery process on the next run? Loop through all the ext-calls that were next-pending and say that these have timed out because they might have been queued?

I see why you're implementing it this way to avoid having to do another persistence inbetween extcall generation and action kickoff, but on first blush it does seem to make the recovery logic a bit more complicated.

crates/lib/vm-driver-thread/src/lib.rs
MOZGIII: Using this would be an akin of calling `exit()` - the unwinding is skipped and all externally owned resources are going to be leaked.
So this is not really intended for any sort of use, but doing an emergency termination of a thread while still keeping some remaining state in order to clean up afterwards.

You are right about the oddities, since we are using the shared tokio runtime - given that its resources would leak as well it is highly likely it'll be corrupted.

So, this is a good find; we should make this call `unsafe`, and perhaps further reduce the API access to the abort handle...

---

UPD: actually, this is all meaningless, and the code is wrong; the semantics I intended was to have a thread control here - but this is not what we have; it's just merely a tokio task - it doesn't expose the under-the-hood thread controls to us here. This should be corrected. Good thing you brought this up.

crates/lib/vm-driver/src/lib.rs
MOZGIII: I imagine this to be as follows:
- we spawn another vm runtime from a snapshot at a state before some effects in question were emitted
- we go though them and emit them again; since always we go off of a deterministic state we can number them, and said numbering will be stable (this is because accepting new external state forces persisting a snapshot before any work can continue)
- we pass the emitted effect to the effect handlers, of which extcalls have a separate persistency layer of their own; it persists the extcall together with its actual sequence number, vm id and the promise state id it resolved/rejects
- the action calls execute after the persistence of an extcall is complete, and has to be done with yet another layer - a reconciler of some sort, that ensures an action is launched somewhere and progresses as long as there is an record for it

This whole part is not actually done yet, and this is what I've had been working on yesterday partially - starting from the interfaces, and it's quite a lot to unpack there.

For this to work, we might actually want to add the extcall numbering to the effector and its interfaces.

piercefreeman: Got it, wasn't clear you were thinking of doing another separate persistence layer in the effect handlers. My initial sense is that should be fine, but might be worth adding a quick comment in the runloop to that effect. ie. we will re-emit events under certain termination criteria so we need extcalls to be aware of their own persistence and recovery semantics.

crates/lib/vm-driver-thread/src/lib.rs
MOZGIII: https://github.com/piercefreeman/waymark/compare/f965d87096d0a3cb4a4493f1a5faeac8ccb94437..702a7f49e906bcd936705218ea9309951527e38b

piercefreeman (approved):

crates/lib/vm-driver/src/lib.rs
MOZGIII: Added the comment to the trait fn; as opposed to documenting specifically the persistence requirements, I documented the idempotence requirement - which is actually more true, and the persistence requirements follow for it for *some* implementations

PR #438, opened 2026-06-23
Add workload pinning subsystem

MOZGIII:
This PR adds the workload pinning subsystem.

In comparison to the old implementation, within the new architecture this subsystem partially replaces the instance queue and locks - while not including any of the actual execution state (i.e. graphs in the old implementation or VM snapshots in the new implementation).

The workload can be anything, but in the current design it is intended to be used for managing the VM runtimes scheduling on the pool of executor nodes (in concrete terms - the `start-workers` instances).

The benefit of this approach is that it decouples the slow and costly operations on the big blobs from the more lightweight data that we can poll often (i.e. the queue and the locking status of a particular workload - or, more concretely - a instance of a VM).

---

This implementation expresses the interaction between the internal control loops in a tight type-safe manner, and also exposes underlying error conditions for explicit error handling.

---

The test coverage is good, and we explicitly cover tricky edge cases, like the behaviour of one loop when the other crashes.

### DIFF
--- a/crates/lib/workload-pinning-manager/src/maintenance/mod.rs
+++ b/crates/lib/workload-pinning-manager/src/maintenance/mod.rs
@@ -1,0 +1,178 @@
+//! Maintenance loop — handles evictions, pinnings refresh, and batch registration.
+//!
+//! # Exit contract
+//!
+//! The maintenance loop keeps running as long as there is work to do.
+//! Work exists whenever **either** condition holds:
+//!
+//! - `batch_rx` is still open — the poll loop is alive and may dispatch
+//!   newly-pinned instance IDs at any time.
+//! - `active_ids` is non-empty — there are in-flight instances whose
+//!   pinnings must be refreshed and whose evictions must be processed.
+//!
+//! The **terminal state** is reached when **both** conditions are false:
+//! `batch_rx` has been closed (the poll loop has exited) **and**
+//! `active_ids` is empty (all in-flight work has been evicted). At that
+//! point no future work can arrive and the loop exits cleanly.
+//!
+//! On exit (whether clean or errored) the remaining `active_ids` set is
+//! sent through `cleanup_tx` so the caller can release any pinnings
+//! that were still held.
+//!
+//! The maintenance loop listens to its own shutdown token for emergency
+//! cancellation — this is separate from the poll loop's graceful-shutdown
+//! token. Under normal operation the loop exits naturally when the poll
+//! loop has stopped and all active IDs have been evicted.
+//!
+//! # Heartbeat retry policy
+//!
+//! Heartbeat refreshes are critical for keeping pinnings alive.
+//! If a refresh fails the loop immediately retries once — this handles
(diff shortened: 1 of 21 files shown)


crates/lib/workload-pinning-manager/src/maintenance/mod.rs
```diff
@@ -0,0 +1,178 @@
...
+                        if !active_ids.is_empty() {
+                            info!("poll loop exited, {} active IDs remain", active_ids.len());
+                        }
+                    }
+                }
+            }
+            Some(id) = evict_rx.recv() => {
+                active_ids.remove(&id);
+                let _ = count_tx.send(active_ids.len());
+                if let Err(error) = (*backend).release_pinnings(node_id.clone(), NEVec::new(id)).await {
```
piercefreeman: It might be worth a comment here that even if this fails, we'll still remove the id from active_ids. So effectively we rely on the TTL expiration to resume this work even during this error.

crates/lib/workload-pinning-manager/src/pinned_handle.rs
```diff
@@ -0,0 +1,48 @@
...
+        Self {
+            id: Some(id),
+            evict_tx,
+        }
+    }
+
+    /// Return a reference to the wrapped instance ID.
+    pub fn id(&self) -> &InstanceId {
+        // SAFETY: `PinnedHandle` is not dropped, so the `id` must be present.
+        unsafe { self.id.as_ref().unwrap_unchecked() }
```
piercefreeman, line 30: Not totally clear why in the lifecycle this must be unsafe.

MOZGIII: `unsafe` block is required because `unwrap_unchecked` fn is unsafe. We don't *have to* use `unwrap_unchecked` here - but we can, and it's cheaper. The whole scope of safety concerns is pretty narrowly encapsulated in this tiny `PinnedHandle` type, so it's unlikely to cause issues during maintenance.

crates/lib/workload-pinning-manager/src/maintenance/mod.rs
MOZGIII: Actually, if we think about it, this is kind of bad; the backend error here will cause the overall loop to terminate - including the heartbeats. The parent loop cleanup will try to clean all the active ids before we exit - except this one.
So, if the cleanup succeeds, we'd be able to quickly re-pin the VMs in a new (or same) node - except for that one that was unlucky to hit the `release_pinnings` fault.

So yeah, this is a good find; also not covered by tests - will add it.

MOZGIII: https://github.com/piercefreeman/waymark/compare/c85a02f2304efa9cf8be484f36df338b9b252ee3..ba5ca30aaa1547a15e67f0758f0e33b51b5fcab0

piercefreeman (approved):
New changes after review good to me!

PR #439, opened 2026-06-24
Add conversions machinery

MOZGIII:
This PR adds a common trait and generally composable system of conversions.

It's like `From`/`Into`, but bound to a third-party converter type as opposed to a source or the target of the conversion.

This gives us the ability to have multiple converters for the same type pairs which do different things - something that really comes in handy given that we have different representations for essentially the same data encoded for actions calls, workflow completions and workflow initializations.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4440,6 +4440,10 @@ dependencies = [
 ]
 
 [[package]]
+name = "waymark-convert-core"
+version = "0.1.0"
+
+[[package]]
 name = "waymark-core-backend"
 version = "0.1.0"
 dependencies = [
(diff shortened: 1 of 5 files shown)


piercefreeman (approved):
This PR as currently scoped seems reasonable - but leaves open the question of future evolution.
Do we have a ticket filed for longer term harmonization across the different uses of values to hopefully reduce the amount of conversion logic we need?

MOZGIII:
We do have https://github.com/piercefreeman/waymark/issues/402

That one above is a broader context issue - but we I'm linking this so it will be mention there; ultimately, I view this not as part of a cleanup, in general, but rather as part of two more concrete projects: 1) making sure transport layer types (proto codegen) are are not used outside of the specific subsystems that directly work with the transport layer; and 2) eliminating redundancies in ways we specify our internal transport protocols (i.e. maybe switch from grpc schema + codegen to source-first schema generation, tbd).

PR #442, opened 2026-06-29
Fix cargo shear

MOZGIII:
This PR correct the warnings at cargo shear.

They don't trigger CI errors, but they show up and clutter the output. There is no option to force them to make the `cargo shear` command to return non-zero exit code afaik, so we should create an issue.


MOZGIII:
nvm, my local carho-shear was outdated

PR #445, opened 2026-07-02
VM effects handling machinery

MOZGIII:
This PR adds machinery required for the VM effects handling.

It covers:
- some adjustments to the existing machinery for conversion
- the universal interfaces for invoking the action calls and collecting the action call results (I have a concrete improvement for this area, that simplified things, but we're moving with this variant for now)
- a transient sleep and action reconcilers - there is NO persistence for actions and sleep for now, which means we don't survive the restart; this is another thing that we'll have to add before we get to production; thankfully this should be easy
- the reconcilers do not yet have proper support for restarts and timeouts - this is something that can be added after the persistence

---

One important details to point out: this is working code, but the implementation needs a rewrite with a better architecture. To that end, this implementation is left in a rather wip state, for instance, the action call correlation parameters (promise state id and effect number) handling is not consistent between the two implementations. The prevailing design should not keep any in-memory maps on the rust side and should pass the metadata through the worker. The worker shouldn't really have a transparent schema for the metadata, since we have a varying variants of the metadata sets (and the corresponding schemas), and it can be fully opaque to the worker.


crates/lib/action-reconciler/src/lib.rs
```diff
@@ -0,0 +1,175 @@
...
+    ActionCallCompletionsProvider: waymark_action_runtime_core::ActionCallCompletionsProvider,
+{
+    /// Wait for the next batch of action-completion settlements.
+    pub async fn poll<Ack>(
+        &mut self,
+    ) -> Option<NEVec<PromiseSettlement<ActionCallCompletionsProvider::Value, Ack>>>
+    where
+        Ack: From<self::Ack>,
+    {
+        let completions = self.provider.wait_for_completions().await.ok()?;
```
piercefreeman: The `.ok()?` here drops the provider error on the floor. Downstream in `extcall-reconciler`'s `get_promise_settlements`, the `Some(...) =` select! pattern treats the resulting `None` as "disable this branch", while the sleep branch keeps waiting on `rx.recv()` — so a fatal provider error (`WorkerPoolGone`, `ChannelClosed`, a resolve failure) never surfaces anywhere. Could we propagate the error out instead of mapping it to `None`?

If the worker pool dies while a workflow has an in-flight action and no pending sleeps, doesn't `get_promise_settlements` just await forever with no log or error?

crates/lib/action-runtime-worker-stream/src/completions_provider.rs
```diff
@@ -0,0 +1,132 @@
...
+impl waymark_action_runtime_core::ActionCallCompletionsProvider
+    for WorkerStreamActionCallCompletionsProvider
+{
+    type Value = waymark_vm_value::ReadyValue;
+    type Error = ReceiveError;
+
+    async fn wait_for_completions(
+        &mut self,
+    ) -> Result<NEVec<ActionCallCompletion<Self::Value>>, Self::Error> {
+        let mut rx = std::mem::replace(
```
piercefreeman: This swaps `self.rx` out for a throwaway closed receiver and only restores it at line 47 on the success path. Since `poll_action_settlements` is raced in a `tokio::select!` against the sleep settler, this future can be dropped while parked in `recv().await` — taking the real receiver with it and leaving `self.rx` as the dummy, so every subsequent call returns `ChannelClosed`. `self.rx.recv().await` / `self.rx.try_recv()` both work directly on `&mut self`, so the swap looks avoidable.

What happens to in-flight action results if a sleep deadline elapses while an action is running and the sleep branch wins the select?

crates/lib/action-runtime-worker-pool/src/completions_provider/direct.rs
```diff
@@ -0,0 +1,57 @@
...
+    ) -> Result<NEVec<ActionCallCompletion<Self::Value>>, Self::Error> {
+        let maybe_completions = self.pool.poll_complete().await;
+        let completions = maybe_completions.ok_or(DirectCompletionsError::WorkerPoolGone)?;
+
+        completions
+            .into_nonempty_iter()
+            .map(|completion| {
+                resolve_completion(&self.correlation_map, completion).map_err(Into::into)
+            })
+            .collect()
```
piercefreeman: `collect::<Result<...>>` short-circuits on the first `ResolveError`, but `resolve_completion` has already removed the correlation-map entries for the completions resolved earlier in the batch — and the batch itself was consumed from `poll_complete`, so those settlements can't be recovered. Same pattern with the `?` in `routed.rs`'s `wait_for_completions`. Resolving per-completion and skipping/logging the failures seems safer than failing the batch.

Since this provider polls the pool with no per-VM filtering, what happens to a good completion that gets drained in the same batch as a stale or foreign token that hits `DispatchTokenNotFound`?

crates/lib/action-runtime-worker-stream/src/completions_provider.rs
```diff
@@ -0,0 +1,132 @@
...
+
+/// Parse an action ID string of the form `"{effect_number}/PromiseStateId({id})"`.
+///
+/// Returns `(EffectNumber(0), PromiseStateId(0))` when parsing fails.
+fn parse_action_id(action_id: &str) -> (EffectNumber, PromiseStateId) {
+    // Find the "/PromiseStateId(" separator
+    let prefix = "/PromiseStateId(";
+    let sep = match action_id.find(prefix) {
+        Some(pos) => pos,
+        None => return (EffectNumber(0), PromiseStateId(0)),
```
piercefreeman: On any parse failure this returns `(EffectNumber(0), PromiseStateId(0))` instead of rejecting the completion — and `PromiseStateId(0)` is a real index, so the vm-driver will resolve/reject the workflow's first promise with this result. The string being parsed is also the `{:?}` Debug output of `PromiseStateId` from `build_dispatch`, so the wire format is coupled to a derive. The bridge path drops unparseable IDs (`action_id.parse().ok()?`); should this do the same, or better, carry the two IDs as structured proto fields?

If a worker echoes back a malformed or legacy-format `action_id`, don't we settle promise 0 with an unrelated action's result?

crates/lib/action-runtime-worker-stream/src/completions_provider.rs
MOZGIII: This code was just nonsense; slipped though the cracks, the interfaces are adjusted such that the code there has `&mut self` access; in essence, we shouldn't need any temporary receiver, as we could just do `self.rx.recv`.

I apologize for calling the review on this level of code quality, I really should not have rushed it. Going to do some internal reviews first next time.

crates/lib/action-runtime-worker-stream/src/completions_provider.rs
MOZGIII: This is a kind-of AI-made dummy implementation; the proper way of doing this would require us to change the worker interface - something that I wanted to do separately. Also, I view this pair of `EffectNumber` and `PromiseStateId` as a particular case of "correlation metadata" - and the problem is sometimes we also have the VM ID that we want to carry around; in the current implementation, this is mostly hardcoded, but I have a branch in the wip that reworks all of this and makes a proper interface for metadata.

---

To address the questions at hand: yes, returning 0 / 0 is never correct, and we should rather fail the execution with an error; I'll adjust the code for that.

crates/lib/action-reconciler/src/lib.rs
MOZGIII: Some general thoughts first:

- The errors have to actually be handled at the `waymark_extcall_reconciler_core::ActionPromiseSettler` implementation; our reconciler is such `waymark_extcall_reconciler_core::ActionPromiseSettler` that wraps `waymark_action_runtime_core::ActionCallCompletionsProvider`.

- The `waymark_action_runtime_core::ActionCallCompletionsProvider` is fallible, while `waymark_extcall_reconciler_core::ActionPromiseSettler` is not; they both have a notion of `Ack` for the settlements/completions, and are supposed to keep them alive and durable persisted up until the settlement is acked - something this *is not yet implemented properly*.

- The infallibility of the `waymark_extcall_reconciler_core::ActionPromiseSettler` is intentional; making a future return `Option` and returning `None` from it is essentially a way to work around returning an infinitely pending promise (`std::future::pending`) without actually doing so, since using that would cause leaks in the `tokio` reactor.

To conclude, this is actually a good find! The issue is at the `extcall-reconciler`: it tries to optimistically keep the promise settlements progressing, even if one of the subsystems dies; which can actually cause the whole thing to hang; rather, it should be pessimistic, and propagate the `None` rather than disabling the select branch.

---

In a more mature implementation the reconcilers would likely have their own loops that would allow us to detect issues and trigger shutdown if anything odd is detected - so the current behavior could be work - but it would be more correct to shutdown the whole promise settler even then. So this is a really good subtle catch.

MOZGIII: Upon closer inspection - the `waymark_extcall_reconciler_core` promise settlers can be made fallible, and that would naturally eliminate all the subtleties and complications.

So, yeah, we'll just propagate the errors as suggested.

crates/lib/action-runtime-worker-pool/src/completions_provider/direct.rs
MOZGIII: Yes, this is a valid correction.

That said, note that the `correlation_map` is a temporary hack and will be gone soon, so this failure will become non-existent. I'll correct it here for now though.

MOZGIII:
@piercefreeman I've addressed the review comments; the implementation is much more resilient to failures now.

piercefreeman (approved):
Nice - thanks for the changes!

MOZGIII:
Pushed some tests corrections, merging once the CI passes

PR #446, opened 2026-07-04
State managers

MOZGIII:
Goes in after #445.

---

This PR adds state managers - a new subsystem that is designed to keep the in-memory objects live while they are used or for a specified liveness period after they are not.

This was originally called "cache", but the "state manager" captures the semantics better.

The subsystem is designed to operate on the "live" objects as opposed to bytes, hence why cache in the naming was discouraged.

---

The state manager itself acts as main the holder of the main volume of complexities; it is well-tested and optimized. The other crates are thin layers that integrate on top, providing the business logic.

The idea is simple - we have an map that keep the values around for the certain keys; the values, when obtained, give out a handle - which, when expires, starts the eviction timer. When the time comes to evict the entry, unless it was obtained again and there's another active handle, we drop the key/value from the map; it can be reconstructed again if needed by fetching it again.

### DIFF
--- a/crates/lib/state-manager/src/lib.rs
+++ b/crates/lib/state-manager/src/lib.rs
@@ -1,0 +1,93 @@
+//! Generic state manager with delayed eviction control.
+//!
+//! See [`State`] and [`Sweeper`].
+
+#![warn(missing_docs)]
+
+mod guard;
+mod handle;
+pub mod provider;
+mod storage;
+mod sweeper;
+
+pub use self::handle::Handle;
+pub use self::provider::Provider;
+pub use self::sweeper::Sweeper;
+
+use core::hash::Hash;
+use std::sync::Arc;
+
+use waymark_nonzero_duration::NonZeroDuration;
+
+use self::guard::Guard;
+use self::storage::Maps;
+
+/// The state that holds `Key`/`Value` pairs.
+///
+/// `Value`s are only exposed as read-only to the consumers.
+///
+/// After the last [`Handle`] to a value is dropped, the value is guaranteed
+/// to remain in the state for a specified retention period, after which
+/// the [`Sweeper`] can pick it up and clear it out.
+///
+/// If the value is not in the state (either it has never been there, or it
+/// was unused and got removed) - a new value is created when obtaining
+/// a [`Handle`]; see [`State::get`].
+pub struct State<Key, Value, Factory> {
+    /// The underlying state maps.
+    maps: Arc<Maps<Key, Value>>,
+
+    /// A factory used to produce the values for this state manager.
+    factory: Factory,
+}
+
+impl<Key, Value, Factory> State<Key, Value, Factory>
+where
+    Key: Eq + Hash,
+{
+    /// Create a new [`State`] and [`Sweeper`] with the specified
+    /// `retention` and `factory`.
+    pub fn new(retention: NonZeroDuration, factory: Factory) -> (Self, Sweeper<Key, Value>) {
+        let maps = Arc::new(Maps::new());
+
+        let sweeper = Sweeper::new(retention, Arc::downgrade(&maps));
+
+        let state = Self { maps, factory };
+
+        (state, sweeper)
+    }
+}
+
+impl<Key, Value, Factory> State<Key, Value, Factory>
+where
+    Key: Eq + Hash + Clone,
+    Factory: waymark_state_manager_core::Factory<Key = Key, Value = Value>,
+    Value: Clone,
+{
+    /// Get or create a new value in the store, and return a [`Handle`] to it.
+    ///
+    /// [`Handle`] provides read-only access to the value, and while
+    /// the [`Handle`] is held the entry will not be removed from the store.
+    ///
+    /// After the last [`Handle`] to the entry is dropped, it is marked for
+    /// eviction, and (unless another [`Handle`] to it is obtained) will
+    /// be removed from the store by the [`Sweeper`] after the corresponding
+    /// retention period.
+    pub async fn get(&self, key: Key) -> Result<Handle<Key, Value>, Factory::Error> {
+        let oncecell = self.maps.acquire(&key);
+
+        // The reference acquired above is owned by `guard` from here on. If the
+        // factory errors or this future is cancelled at the `.await` below,
+        // dropping `guard` releases the reference; on success `guard` moves
+        // into the returned `Handle` and releases it when the last handle is
+        // dropped.
+        let guard = Guard::new(Arc::downgrade(&self.maps), key);
+
+        let value = oncecell
+            .get_or_try_init(|| self.factory.produce(guard.key()))
+            .await?
+            .clone();
+
+        Ok(Handle::new(guard, value))
+    }
+}
--- a/crates/lib/state-vm-runtimes/src/spawner.rs
+++ b/crates/lib/state-vm-runtimes/src/spawner.rs
@@ -1,0 +1,158 @@
+//! VM driver thread spawning.
+
+use std::sync::Arc;
+
+use tokio_util::sync::CancellationToken;
+use tracing::Instrument as _;
+
+/// A spawned VM handle that exposes the access to the VM.
+#[must_use = "driver thread is aborted when this handle is dropped"]
+pub struct Spawned {
+    /// Gracefully cancels the driver loop.
+    driver_loop_cancel: CancellationToken,
+
+    /// A handle to allow joining on the driver completion task.
+    ///
+    /// That's the one that waits for the driver to complete.
+    driver_completion_task: Option<tokio::task::JoinHandle<()>>,
+
+    /// Cancelled when the driver thread exits (the VM is evicted).
+    driver_evicted: CancellationToken,
+
+    /// Opaque handles kept alive while this VM is running.
+    #[allow(
+        unused,
+        reason = "these handles are kept around for the drop semantics, \
+                    they don't actually have to be used"
+    )]
+    keepalive_handles: Vec<Box<dyn Send + Sync>>,
+}
+
+impl Spawned {
+    /// Returns a future that resolves when the VM driver thread exits
+    /// (the VM is evicted).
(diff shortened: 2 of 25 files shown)


crates/lib/state-manager/src/lib.rs
```diff
@@ -0,0 +1,314 @@
...
+                    oncecell
+                }
+            }
+        };
+
+        self.maps.pending_evictions.remove(&key);
+
+        let value = oncecell
+            .get_or_try_init(|| self.factory.produce(&key))
+            .await?;
```
piercefreeman: `refs` is incremented (lines 204/211) before this await, but the only decrement lives in `Handle::drop` — so the `?` here, or the caller dropping the `get()` future at this await point (timeout/select), exits without ever constructing a `Handle`. The entry's refs stays elevated forever: it can never hit 0, never enters `pending_evictions`, and `remove_if(.., refs == 0)` can never sweep it. Every retry of a failing key leaks another ref. A small drop-guard that decrements refs unless the Handle was built would cover both paths. Worth noting the integration tests all use `Error = Infallible` factories, so neither path is exercised.

If `produce` fails once with a transient backend error and the caller retries, can that key ever be evicted afterwards — and when it later succeeds, what unpins the VM driver thread it now holds live?

crates/lib/state-vm-runtimes/src/spawner.rs
```diff
@@ -0,0 +1,152 @@
...
+            Ok(()) => {}                         // fallthrough to disable abort-on-drop
+            Err(err) if err.is_cancelled() => {} // fallthrough to disable abort-on-drop
+            Err(err) => std::panic::resume_unwind(err.into_panic()), // unwind and trigger abort-on-drop
+        }
+    }
+}
+
+impl Drop for Spawned {
+    fn drop(&mut self) {
+        self.driver_loop_cancel.cancel();
```
piercefreeman, line 72: Eviction is fire-and-forget: this only cancels the token and returns — it neither waits for the driver to exit (`shutdown()`/`evicted()` exist but nothing wires them into eviction) nor gives the driver a chance to persist. The driver loop checks cancellation at the top of the iteration, *before* the persist step, and returns `Err(Cancelled)` with no final snapshot. So state produced by an already-executed `handle_effect` (side effect visible externally) is discarded, and the next `get(vm_id)` revives from the pre-effect snapshot. Separately, since the sweep removes the entry before the eviction callback runs and the thread isn't joined, a re-fetch can run `load_for_revive` while the old driver's in-flight `store_snapshot` is still completing — and `StoreSnapshot` has no version/CAS guard to reject the late write.

If a VM is swept right after `handle_effect` returns but before the next loop iteration persists, doesn't the revived VM re-execute an effect that already ran externally?

crates/lib/state-manager/src/lib.rs
```diff
@@ -0,0 +1,314 @@
...
+        maps.pending_evictions.retain(|key, orphaned_since| {
+            let retain = now.duration_since(*orphaned_since) <= retention;
+            if !retain {
+                cleanup_queue.push((*key).clone())
+            }
+            retain
+        });
+
+        for key in cleanup_queue {
+            let Some((key, entry)) = maps.entries.remove_if(&key, |_, entry| entry.refs == 0)
```
piercefreeman: This predicate checks only `refs == 0`, not whether the key re-entered `pending_evictions` with a fresh timestamp after the `retain()` above snapshotted the cleanup queue. So: sweep expires K into the queue → concurrent `get(K)` bumps refs (its `pending_evictions.remove` is a no-op, K is already stripped) → the Handle drops a moment later, refs back to 0 with a fresh timestamp → this `remove_if` still succeeds and evicts K immediately — retention window ~0 instead of the configured duration, which for an `Arc<Spawned>` value tears down the VM right after use. The long comment below narrates this exact interleaving but only assesses the benign None-branch as harmless; step 4 is the retention violation. Folding the timestamp into the entry itself (`orphaned_since: Option<Instant>` checked inside this predicate) would make eviction eligibility atomic and remove the cross-map race class entirely.

If a handle is acquired and dropped while a sweep is mid-flight, what stops the just-released value from being evicted milliseconds after its last use?

crates/lib/state-vm-runtimes/src/spawner.rs
```diff
@@ -0,0 +1,152 @@
...
+                match error {
+                    waymark_vm_driver_thread::Error::Driver(error) => {
+                        tracing::error!(?error, "vm driver terminated");
+                    }
+                    waymark_vm_driver_thread::Error::Thread(error) => {
+                        tracing::error!(?error, "vm thread panicked");
+                    }
+                }
+                tracing::debug!("vm driver thread exited");
+                evicted.notify_waiters();
```
piercefreeman: `notify_waiters()` wakes only `Notified` futures that are already registered and stores no permit, but `evicted()` mints a fresh `Notified` per call — so the doc contract ("resolves when the VM driver thread exits") doesn't hold for any caller that starts waiting after (or races with) driver exit; that future pends forever. A latched primitive is needed here — `CancellationToken` is already imported in this file (or a `watch` channel) and resolves immediately for late subscribers.

If the driver has already exited when a caller awaits `evicted()`, what wakes that future?

crates/lib/state-manager/src/lib.rs
MOZGIII: Oof, that's a bad one; yeah, good catch!

I've corrected this issue. Didn't use a drop guard, but rather added a manual error check. Also added a test to cover this case.

crates/lib/state-vm-runtimes/src/spawner.rs
MOZGIII: There are three issues here:

1. The question that you've brought up:

    > If a VM is swept right after `handle_effect` returns but before the next loop iteration persists, doesn't the revived VM re-execute an effect that already ran externally?
    
    Yes, that's precisely what's going to happen; the VM will reach the same effect, emit it again, and the `handle_effect` will be called; the proper effect handler implementation is supposed to durably persist and de-duplicate the call. In out architecture, this is the role of the reconcilers - which are, at this point, pretty much, almost direct wrappers to the low level action executors. This is temporary, and in the proper production design they'll touch the database for every action call, before triggering it again.

    This kind of integration will, most likely, require us to expose somewhat different knobs to the user - i.e. in addition to timeout and retries it should be possible to choose how to reconcile the action call - prefer at-most-once or at-least-one execution. So, this is postponed until we can take a deeper look into each of the subsystems.

    I was mostly focusing on the global architecture (core interfaces, abstractions and composability in the system) with certain known things left out for now.

2. Race condition between the VM eviction (i.e. final snapshot) and the re-loading it for more work.

    This is closely related to a fundamental issue - there will always be a possibility that the process dies while the VM is executing, and we fail to emit the snapshot; so, the same solution would ultimately apply here - we just accept that effects can be triggered twice, and reconcile the real side effects against what's been requested by the VM. Pretty much, see above.
    
    But in this particular case - we can't really wait for termination (no async drop in rust), so we do a best-effort hint to the driver to stop doing more work. In the contexts where we have a way to properly block - there's a `shutdown` method that allows it.

    For instance: in the subsequent PRs we would have two handles - the vm-runtime-state handle and workload-pinning handle - and  we'd want to shutdown the VM first by consuming the vm-runtime-state, then drop the pinning handle.

    This is something I always intended, but the current API doesn't allow (cause `shutdown` takes `self` by-value) - yet another overlook from my part; to solve this, I'm adding a new method - `Spawned::trigger_eviction(&self)` - that will just trigger the cancellation of the driver loop. We already have the `evicted` method to allow waiting for the cancellation - so this is what we'd use in the subsequent PRs case from above.

3. No final snapshot

    A final snapshot before the eviction is something that's possible to do, but something that I've decided to skip at the time when I started working on the design. The right place for it would, most likely, be at the vm driver loop, since it is what does the persistence. But it's not there, precisely because there are big uncertainties with related to race conditions. In particular, when the cancellation is triggered, it is unclear whether we are still holding the workload pin for this VM or not (in some cases - like when the drop is triggered - we don't); so the easier thing to do, design wise, was to not do any "final snapshot" at all.

    This is something we can add later, or, perhaps, even now. But I'd postpone for later to have a bigger picture first. I'm thinking, we should only to the final snapshot in the flow that *guarantees* we're in a situation where we're still holding the workflow pin. 

    Maybe the vm driver should return the VM runtime alongside with all the errors - just to give us a chance to do something with it, like a final snapshot...

crates/lib/state-vm-runtimes/src/spawner.rs
MOZGIII: Huh, right, I missed this one too

crates/lib/state-manager/src/lib.rs
MOZGIII: This is something I though I tackled thoroughly, but I called it done too soon... I thought about whether we need the generation key - and it seemed like we don't, but we clearly do need it.

I've added this as suggested.

My only concern is the code is now kind of tempting to be reimplemented to iterate over just the main `map` - but that would be incorrect, since the whole point of having the second `pending_evictions` is to avoid iterating through he main `map` - but it now seemingly have all the data to implement everything without the second map. This is doable, but it should be way worse based on the rough assumptions (i.e. create much worse locking profile). This is just speculation of course, since we don't have real load tests or even synthetic benchmarks.

crates/lib/state-vm-runtimes/src/spawner.rs
MOZGIII: I've swapped Notified with the cancellation token - this is the right call.

crates/lib/state-manager/src/lib.rs
piercefreeman: Happy to roll with the current logic.

MOZGIII:
I've updated this PR with some more changes to address the newly discovered issues with correctness - and to land some general code quality improvements that I've had in mind.
I think it should be ready for merge now...

piercefreeman (approved):

PR #448, opened 2026-07-04
Implement vm-runtimes and workload pinning backends for postgres

MOZGIII:
Goes in after #447.

---

This PR implements a full backend for workload pinnings and vm runtimes - both for submitting the initial workload into the pipeline, and for "revival" of the vm after it has been evicted.

---

As with the overall architecture, the database model in denormalized a bit to provide a better layout for our operations. Polling happens only on the workload pinnings, and they don't have any heavy columns - like the snapshot blobs.

### DIFF
--- a/crates/lib/backend-postgres/src/vm_runtimes/mod.rs
+++ b/crates/lib/backend-postgres/src/vm_runtimes/mod.rs
@@ -1,0 +1,93 @@
+//! Postgres backend for VM runtime snapshot persistence.
+
+pub mod error;
+
+#[cfg(test)]
+mod tests;
+
+use sqlx::Row;
+use waymark_ids::{InstanceId, WorkflowVersionId};
+use waymark_observability::obs;
+use waymark_timed_future::TimedFutureExt as _;
+
+use crate::PostgresBackend;
+
+impl waymark_state_vm_runtimes_backend::HasVmId for PostgresBackend {
+    type VmId = InstanceId;
+}
+
+impl waymark_state_vm_runtimes_backend::HasExecutableId for PostgresBackend {
+    type ExecutableId = WorkflowVersionId;
+}
+
+impl waymark_state_vm_runtimes_backend::StoreSnapshot for PostgresBackend {
+    type Error = error::StoreSnapshotError;
+
+    #[obs]
+    #[function_name::named]
+    async fn store_snapshot<'a>(
+        &'a self,
+        vm_id: &'a InstanceId,
+        data: &'a [u8],
+    ) -> Result<(), Self::Error> {
+        Self::count_query(&self.query_counts, "update:vm_runtime_snapshots_snapshot");
+        let result = sqlx::query(
+            r#"
+            UPDATE vm_runtime_snapshots
+            SET snapshot = $2, updated_at = NOW()
+            WHERE vm_id = $1
+            "#,
+        )
+        .bind(vm_id)
+        .bind(data)
+        .execute(&self.pool)
+        .timed(crate::query_timing_histogram!(
+            "update:vm_runtime_snapshots_snapshot"
+        ))
+        .await
+        .map_err(error::StoreSnapshotError::Sqlx)?;
+
+        if result.rows_affected() == 0 {
+            return Err(error::StoreSnapshotError::NotRegistered(*vm_id));
+        }
+
+        Ok(())
+    }
+}
+
+impl waymark_state_vm_runtimes_backend::LoadForRevive for PostgresBackend {
+    type Error = error::LoadForReviveError;
+
+    #[obs]
+    #[function_name::named]
+    async fn load_for_revive<'a>(
+        &'a self,
+        vm_id: &'a InstanceId,
+    ) -> Result<waymark_state_vm_runtimes_backend::RevivePayload<WorkflowVersionId>, Self::Error>
+    {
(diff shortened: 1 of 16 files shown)


crates/lib/backend-postgres/src/vm_runtimes/mod.rs
```diff
@@ -0,0 +1,91 @@
...
+        &'a self,
+        vm_id: &'a InstanceId,
+        data: &'a [u8],
+    ) -> Result<(), Self::Error> {
+        Self::count_query(&self.query_counts, "update:vm_runtime_snapshots_snapshot");
+        let result = sqlx::query(
+            r#"
+            UPDATE vm_runtime_snapshots
+            SET snapshot = $2, updated_at = NOW()
+            WHERE vm_id = $1
```
piercefreeman, line 38: This UPDATE is keyed only on `vm_id` — it doesn't consult `workload_pinnings` and there's no version column on the row. What happens if a node stalls (GC pause / partition), its lease expires, the VM gets re-polled and revived on another node, and then the old node's in-flight `store_snapshot` lands? Doesn't it silently overwrite the new owner's newer snapshot? Wondering if this needs the caller's node id plus an `AND EXISTS (... node_id = $me AND expires_at > NOW())` guard, or a CAS/version column.

crates/lib/backend-postgres/src/workload_pinning/mod.rs
```diff
@@ -0,0 +1,177 @@
...
+    > + Send
+    + 'a {
+        let ids: NEVec<InstanceId> = instance_ids.into_nonempty_iter().collect();
+        async move {
+            Self::count_query(&self.query_counts, "update:workload_pinnings_refresh");
+            let rows = sqlx::query(
+                r#"
+                UPDATE workload_pinnings
+                SET expires_at = $3, updated_at = NOW()
+                WHERE node_id = $1 AND instance_id = ANY($2) AND expires_at > NOW()
```
piercefreeman: Refresh checks expiry with the database clock (`NOW()`, and `now` is ignored at the signature), while `poll_unlocked` above checks `expires_at <= $1` with the caller's clock — and the `expires_at` values themselves come from whichever node built the `Pinning`. What happens when one node's clock runs a bit fast — can it steal a live lease through poll while the owner's refresh still succeeds by DB time? Would it be simpler to make Postgres the single clock (use `NOW()` in the poll predicate too, maybe even compute `expires_at = NOW() + $interval` server-side)?

crates/lib/backend-postgres/src/vm_runtimes/mod.rs
MOZGIII: I have thought about this too.

The way the abstractions are set up right now the snapshot persister has no notion of the node id. This can be adjusted if needed, so it's a minor thing.

More serious thing is this could be costly performance-wise, but it really shouldn't ever be needed. The other logic has to be set up in such a way that it's never possible to issue a snapshot write unless we hold an active pinning; we can, of course, do this - but having *just the reliable shared locks should be enough* to *guarantee* the safety of these writes, at least in theory. So, if we do this - I'd consider it a temporary solution to prevent potential early-development troubles. Not too good at that really - cause the snapshot state corruption is really not the main issue here - it's concurrently executing actions / side effects that will be the most harmful. So we really can't let the concurrent execution of the VMs happen - and thus the corrupted snapshots also shouldn't *really* be a big issue.

In practice, if we have two concurrently executing VMs we'll most likely see their effects deduplication not working properly - since the quicker VM would ack the completions and clean them up. I'm thinking, this completion clean-up is where we might want to add another layer of synchronization - i.e. the completion for a VM can only be acked by the same node that holds the workload pin for the VM; there it'd make more sense since the clean-ups will always happen in bulk...

---

Anyhow, I don't know if this is actually needed or not just yet. It was left out intentionally, but the could be reasons to add it later.

crates/lib/backend-postgres/src/workload_pinning/mod.rs
MOZGIII: Right, the `updated_at = NOW()` is correct, but the `expires_at > NOW()` is not; we can't mark them as separate types in postgres (can we?) - but really the `updated_at` is supposed to be server time, while `expires_at` is supposed to always come from the client.

The `_now` in the signature hits this is a mistake, which is absolutely right.

MOZGIII: On the topic of using the actual `NOW()` for clocks everywhere - we kind of need a way to guarantee that the nodes themselves don't have time skews, otherwise pretty much anything we do with the time is incorrect.

> What happens when one node's clock runs a bit fast — can it steal a live lease through poll while the owner's refresh still succeeds by DB time?

For that to happen, the clock would actually have to run substantially quicker, and measurably so - because we don't refresh the pins just in time for them to expire - we should normally have like up to 1/3-ish of the expiration timeout set for the heartbeat interval... I don't know what would be a typical production value for the expiration timeout - but with a clock skew of 5-7 mins we'll start getting TLS errors when connecting to the HTTPS websites.

We could use `NOW()` - but the problem then is the shared state has a different view of the time than any of the nodes; in a way, this just adds one more node (DB) to the equation. I'd imagine the workers execution environment to often be more homogeneous among just each other than with the DB.

piercefreeman (approved):

PR #474, opened 2026-07-17
Parking support at the workload pinnings

MOZGIII:
Goes in after #473.

---

This PR adds the support for parking the workloads, removing them for being candidates for pinning (aka from the runnable set).

Parked workloads can later be unparked in order to resume their execution.

---

As of now, we don't have parking for VMs blocked to promises or unparking support for the actions completions or sleep timers; we also don't yet have durable sleep - and that's what's going to be added next. After, we'll add support for the unparking and, finally, for the parking of the VM that are blocked on promises for longer than a certain period of time.


crates/lib/workload-pinning-core/src/lib.rs
```diff
@@ -0,0 +1,15 @@
+//! Core types for workload pinning.
+
+#![warn(missing_docs)]
+
+/// What happens to a workload when it is unpinned.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum UnpinMode {
+    /// End the pinning but keep the workload runnable: it may be pinned
+    /// again by any node.
```
piercefreeman, line 31: Would be good to add a bit more detail here, like: this is appropriate for any situation where another attempt might make progress.

crates/lib/workload-pinning-core/src/lib.rs
```diff
@@ -0,0 +1,15 @@
...
+    Release,
+
+    /// End the pinning and remove the workload from the runnable set:
+    /// it stays unpinnable until it is made runnable again.
```
piercefreeman: Similarly: "made runnable again by a deliberate operation".

piercefreeman (approved):

crates/lib/workload-pinning-core/src/lib.rs
MOZGIII: Actually it was documented like what you propose before, but it's not actually valid; I wasn't able to find a cleaner criteria than just "a negation of when park is justified"; in particular - it's not always appropriate to release, since sometimes it'll cause re-pin spinloop; at the same time - parking is not always appropriate either, it'll cause liveness loss unless properly justified.

In fact, neither of the operation is valid without explicitly checking for the spinpools and/or the ordering against unpark races, and this is one of the weak spots in the design (checking far-away invariants is always tricky, ideally we'd find a way to guarantee impossibility via a type-system somehow, but I don't think we can here). I guess *this* is what I failed to express in these original attempts to document and capture the design, but I sure can do it now.

PR #476, opened 2026-07-19
Durable sleep

MOZGIII:
Goes in after #475.

---

This PR implement durable sleep. Unlike action calls, sleep is a simpler matter of having a certain time polled as being passed - so it doesn't have two separate storages for requesting and completion - just one path to request a wake-up after a certain time has passed; the request is removed when a VM acknowledges it, eliminating the need for a separate clean-up.


piercefreeman (approved):
The crux of this design - just for posterity - is that `sleep_requests` can keep track of the pending sleep. But nothing needs to update the row at that moment that it's due (wake_at <= now) because we can just use promise settlement to applied and persist. At which point acknowledgement deletes it.

PR #485, opened 2026-07-21
Add support for racing promises at the vm runtime

MOZGIII:
Goes in after #484.

Superseded by #486.

---

This PR adds support for racing promises and obtaining an indicator of which of the promises finished first.

It theoretically weakens the determinism guarantees of the VM runtime, however in practice we have implemented the vm-driver in such a way that the execution is still deterministic with respect to the snapshots history in relation to the promise resolutions; this means that the determinism depends on the order at which the database and rust-side internals produce the promise settlements matters.

This is important for future features like VCR; in essence this means that the level at which we have to record the VCR logs (tapes) is the promise settlements so we can record the exact same effective order those records appear (and preserve the VM runtime `run()`s / promise settlement batching lock-steps). This shall make the VCR feature capture all non-deterministic behaviors such that it can faithfully reproduce the life of a recorded VM on replay. Todo: verify this statement.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5913,6 +5913,7 @@ name = "waymark-vm-interpreter-coreset"
 version = "0.1.0"
 dependencies = [
  "derive-where",
+ "nonempty-collections",
  "thiserror",
  "waymark-vm-bytecode",
  "waymark-vm-exception-handler",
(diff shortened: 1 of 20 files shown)


MOZGIII:
Superseded by https://github.com/piercefreeman/waymark/pull/486.

Ultimately, implementing races at the control flow level is a better fit for us than at a runtime value level.

PR #498, opened 2026-07-23
Upgrade integration-test binary with the new VM

MOZGIII:
Goes in after #497

---

This PR updates the integration-test binary with the new VM. In-memory and postgres backends variation is replaced with transient and durable execution.

Some additional refactoring is applied on top to split the code into mods. The PR can be inspected per-commit for more obvious diff comparison (the update is performed inline first, add then split into mods - so check the first commit when evaluating the port correctness verification).


piercefreeman (approved):

crates/bin/integration-test/src/durable.rs
```diff
@@ -0,0 +1,259 @@
...
+        vm_sweeper,
+        durable_action_completions_writer,
+        durable_action_completions_poller,
+        durable_action_completions_acker,
+        durable_sleeps_poller,
+        durable_sleeps_acker,
+        action_effect_reconciler_lock_renewal,
+    } = handles;
+
+    let _ = tokio::time::timeout(Duration::from_secs(5), pinning_manager).await;
```
piercefreeman, line 174: Should we retain the JoinHandle and abort it if this timeout elapses? Passing it by value means timeout drops the handle on expiry, which detaches the still-running task; the later worker-pool teardown can race it.

piercefreeman: Could shutdown and joining live as one method on waymark_execution_bringup::Handles? It could own the timeouts and abort stragglers once, while benchmark, integration, and start-workers all call the same teardown path.

MOZGIII: This shutdown logic is a placeholder - it's actually not even entirely correct, and we'd want to expose a safer global shutdown / task management logic with three-state shutdown; the idea is we'd have three phases: graceful-shutdown (things are let to complete gracefully however long it takes), force-shutdown (things don't wait for high-level in-flight operations to complete, but the finalization routines run to completion) and emergency-shutdown (things are aborted immediately without being given a change to run async code to completion; this means locks are still held, but the os descriptors can close properly); this complements the os-level `sigkill` to get four ways to shut the program down.

The way it relates to this particular code is we'd rather have a single task manager that we register handles on that dictates this unified logic to be applied to all the tasks, as opposed to managing each one of them explicitly.

This new bringup logic is compatible with this design, but the worker pools still have some rough edges; I refactored the part that manages the processes some time ago to make it safe - but we still need to clean up the part that exposes then pool (the senders) to the VM execution driver (a minor concern) and the part that talks to the actual workers (so we can propagate the cancellation - this is also a pre-requisite for a proper shutdown management flow).

---

You are right, placing this logic in a helper fn is an easy way to avoid code-duplication, but at the same time it hides important implementation details that are non-trivial - I usually leave them exposed in such cases.

I think here it would make sense though, since I've contemplated the same thing and decided against it - but I guess it's a more visible concern that I've anticipated...

PR #500, opened 2026-07-23
Upgrade fuzzer binary with the new VM

MOZGIII:
Goes in after #499


piercefreeman (approved):
I haven't the fuzzer deliver meaningful insights in the past - so might be worth just jettisoning it from the codebase for now to clean things up. But otherwise this port looks fine to me.

PR #504, opened 2026-07-23
Add support for multi-value assignments

MOZGIII:
Goes in after #502


piercefreeman (approved):

crates/lib/vm-compiler-for-ast-old/src/function/compiler/lowering/assignment.rs
```diff
@@ -113,6 +121,87 @@ where
...
+        self.context.emitter.switch_to(unpack_state);
+        for (item_index, target) in targets.into_iter().enumerate() {
+            let item_index = i64::try_from(item_index).expect("target count fits in i64");
+            let index_value = Lowering::lower_literal(&Literal::Int(item_index))
+                .map_err(Error::LiteralLowering)?;
+            let index_register = self.context.local_frame.allocate_register();
+            self.context
+                .emitter
+                .emit_load_const(index_register, index_value);
+            self.context.emitter.emit_index(
```
piercefreeman, line 194: Could the RHS alias the first target here? For a, b = a, a variable RHS reuses a’s register, so the first emit_index overwrites the source before index 1 is read. Should we copy the RHS or extract every item into temporaries first?

MOZGIII: Yes, this is a problem; current architecture is kind of primitive, there's no comprehensive aliasing analysis facility; I'll fix it with an ad-hoc fix... But this is not scalable.

That said - fine for now, given our plans I guess.

PR #505, opened 2026-07-23
Implement range assignments at the vm compiler

MOZGIII:
Goes in after #504

---

This PR adds `range` assignment support via lowering it into a loop that populates an array

### DIFF
--- a/crates/lib/vm-compiler-for-ast-old/src/function/compiler/plan/assignment.rs
+++ b/crates/lib/vm-compiler-for-ast-old/src/function/compiler/plan/assignment.rs
@@ -1,7 +1,7 @@
 //! Assignment planning.
 
 use nonempty_collections::NESlice;
-use waymark_vm_ast_old::{ActionCall, Expr, Spanned};
+use waymark_vm_ast_old::{ActionCall, Expr, FunctionCall, GlobalFunction, Spanned};
 
 use super::ErrorFor;
 use super::parallel::ParallelAssignmentPlan;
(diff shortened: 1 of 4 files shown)


piercefreeman (approved):

crates/lib/vm-compiler-for-ast-old/src/function/compiler/plan/assignment.rs
```diff
@@ -66,6 +66,15 @@ where
         /// Expression evaluated once and unpacked into the targets by index.
         value: &'a Spanned<Expr>,
     },
+
+    /// An assignment materializing a `range(...)` call into a list.
+    RangeValues {
```
piercefreeman, line 71: Could range be lowered once as a normal value expression instead of adding an assignment-only RangeValues plan? Then values = range(...), return range(...), and nested or call-argument uses would all share the same implementation.

MOZGIII: Yes, this is possible. I kind of didn't want to add a range value type at the runtime to keep it minimal, and to see how can we push it without bloating the value types.

If we had a native range value type, we could make native iterator instruction on top of it. My concern is the composition and scope of these instructions (and the bloat) in the context of having multiple-language instruction sets that have to be mixed and matched together. Doing it this way is just a bit simpler for now.

I'd address this after we also have a proper HIR / LIR with a simpler and more unified codepaths and explicit coupling from the frontend compilers. Right now the coupling is implicit - i.e. we only support Python and that's it - but if we were to express that a compiler needs something language-specific - like certain way of representing "iteratables" - and have that notion shared across compiler and corresponding VM runtime - that would be much easier to land.

That said - it's not a hard blocker; the current way of doing things just pushes complexity away from the code part that we'll likely to work on in the near future, in a way that's valid (afaik python did this kind of unwinding too up until relatively recently) and compartmentalized.

PR #511, opened 2026-07-24
Add batcher

MOZGIII:
Goes in after #510 

---

This PR adds batcher - a buffering / flushing primitive to enable efficient batch operations at the cost of potential added latency.


piercefreeman (approved):
Nice abstraction

PR #512, opened 2026-07-24
Add JavaScript and Next.js workflows to the VM

piercefreeman:
## Summary

- add a repository-root npm workspace for javascript/packages/* with one lockfile and root build, test, and lint orchestration
- run JavaScript tests as an independent CI gate and require them before package releases
- add explicit Python and JavaScript action runtime identity across the VM, wire protocol, bridge, and workers
- add the typed action(async function ...) and Workflow.run(...) TypeScript API, compiler, exact-byte executable hashing, source spans, and code-frame diagnostics
- add Next.js build and transient execution adapters that delegate resolution through the bundler callback
- add an opt-in Node worker and runtime-routed durable execution path without extending the deprecated DAG compiler
- document the supported syntax and deployment model, with a durable Postgres VM integration test

## Planned Commit stack

1. Add explicit action runtime identity
2. Add TypeScript wire and value boundary
3. Compile TypeScript workflows for the VM
4. Add transient execution and Next build adapter
5. Add JavaScript worker bridge client
6. Route VM actions to JavaScript workers
7. Document and verify JavaScript VM support
8. Add repository JavaScript workspace
9. Run JavaScript tests in CI


crates/bin/start-workers/src/main.rs
```diff
@@ -1,10 +1,10 @@
-//! Start Workers - Runs the core runloop with Python worker pool.
+//! Start Workers - Runs the VM execution subsystem with action worker pools.
 //!
 //! This binary starts the worker infrastructure:
 //! - Connects to the database
 //! - Starts the WorkerBridge gRPC server for worker connections
-//! - Spawns a pool of Python workers
-//! - Runs the core runloop to process queued workflow instances
+//! - Spawns Python workers and, when configured, JavaScript workers
```
piercefreeman: JS should be at the same semantic level as when we reference Python. Should be "Spawn supported language workers"

crates/bin/start-workers/src/main.rs
```diff
@@ -13,6 +13,10 @@
 //! - WAYMARK_USER_MODULE: Python module(s) to preload (comma-separated)
```
piercefreeman: Should also be standardized to WAYMARK_PYTHON_USER_MODULE.

No need for backwards compatibility since we're cutting a new major release with our changes.

We should also have a new parameter to select from the list of worker languages in `ActionRuntime`: python or javascript... one `start-workers` executable should only own one for simplicity of handling the brokers.

crates/bin/start-workers/src/main.rs
```diff
@@ -112,7 +120,40 @@ async fn main() -> Result<()> {
     )
     .await?;
 
-    let process_pool = Arc::new(process_pool);
+    let python_process_pool = Arc::new(python_process_pool);
+
+    // JavaScript is opt-in: both count and generated bundle are required.
+    let (javascript_process_pool, javascript_bridge_task) =
```
piercefreeman: Refactor python into its own bringup function in this file and same with the js specific code... it'll be useful in the future when we have more languages under support.

crates/lib/action-runtime-worker-stream/src/requester.rs
```diff
@@ -91,3 +100,41 @@ fn build_dispatch<Metadata: Encode>(
...
+    use super::*;
+
+    struct EmptyMetadata;
+
+    impl Encode for EmptyMetadata {
+        fn encode(&self, _writer: &mut Vec<u8>) {}
+    }
+
+    #[test]
+    fn dispatch_preserves_javascript_runtime() {
```
piercefreeman: While technically yes, this test preserves the JS runtime, it's really testing a deeper fact that this logic preserves any runtime. I'd suggest changing the title and we can just use JS as an example of one of these runtimes.

crates/lib/execution-bringup/tests/javascript_postgres.rs
```diff
@@ -0,0 +1,276 @@
...
+fn main(input: [], output: [result]):
+    result = @actions.greet(name="Ada")
+    return result
+"#
+        .trim(),
+    )
+    .expect("parse test IR");
+
+    // The textual IR predates runtime annotations and deliberately defaults
+    // actions to Python. This test exercises the JavaScript source-AST path,
```
piercefreeman: Per our previous comment, we should no longer default actions to Python... this means that we'll potentially have to do this manipulation logic in a lot more places (tests mostly I imagine). I suggest you add a lib create that adds utility functions for looking up actions and statements in order to check what kind of runtime provider they require - as well as functions to allow easier modifications of those definitions so we don't need this kind of modification adhoc in other places.

crates/lib/execution-bringup/Cargo.toml
```diff
@@ -36,5 +36,23 @@ tokio = { workspace = true, features = ["sync", "macros", "rt"] }
...
+waymark-workflow-completion-core = { workspace = true }
+waymark-workflow-service-vm-executables = { workspace = true }
+waymark-workflow-service-vm-runtimes = { workspace = true }
+
+nonempty-collections = { workspace = true }
+serde_json = { workspace = true }
+serial_test = { workspace = true }
+sqlx = { workspace = true }
+uuid = { workspace = true }
+
```
piercefreeman: Not sure why our tests require all these additional dependencies where they didn't before; seems like you're increasing the scope of tests to cover more than they're intended to cover given the scope of execution bringup.

crates/lib/vm-compiler-for-ast-old-action-ref/src/lib.rs
```diff
@@ -53,3 +54,26 @@ pub fn lower_action_ref(call: &ActionCall) -> ActionRef {
         exception_types,
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn preserves_javascript_runtime() {
```
piercefreeman: You have too many tests that are testing the same thing; I think you can remove most of these "perserve" tests if you just enforce that runtime doesn't have a default and therefore needs to be explicitly configured and passed.

crates/bin/start-workers/src/main.rs
piercefreeman: Updated the binary overview to “Spawns supported language workers” and removed the Python/JavaScript asymmetry from the surrounding worker lifecycle wording.

crates/bin/start-workers/src/main.rs
piercefreeman: Reworked configuration around a required WAYMARK_ACTION_RUNTIME parsed as ActionRuntime. Python now uses WAYMARK_PYTHON_USER_MODULE with no compatibility alias. Each start-workers process creates exactly one selected language pool and one bridge, using the shared count/address settings.

crates/bin/start-workers/src/main.rs
piercefreeman: Split language-specific startup into run_python_workers and run_javascript_workers. The common runloop/status/webapp lifecycle remains in run_worker_runtime, so adding a language does not add another conditional pool to the process.

crates/lib/action-runtime-worker-stream/src/requester.rs
piercefreeman: Renamed this to dispatch_encodes_the_declared_runtime. JavaScript remains the example value, while the test name now states the general serialization invariant.

crates/lib/execution-bringup/tests/javascript_postgres.rs
piercefreeman: Removed the implicit Python default: raw IR actions begin Unspecified, IRParser requires an ActionRuntime, and every parser/helper caller passes it explicitly. Added waymark-vm-ast-old-proto-helpers for recursive action lookup, required-runtime inspection, and safe program/expression runtime mutation across every AST variant.

crates/lib/execution-bringup/Cargo.toml
piercefreeman: Deleted the broad javascript_postgres execution-bringup test and removed all 18 dev-dependencies it introduced. Runtime behavior is now covered at the owning compiler, protocol, AST conversion, and worker-pool layers.

crates/lib/vm-compiler-for-ast-old-action-ref/src/lib.rs
piercefreeman: Made runtime structurally explicit instead of adding more preservation checks: parser entry points and every old-AST action constructor now require ActionRuntime, and unspecified protobuf actions remain errors. Consolidated the action-ref check into the existing full metadata/policy lowering invariant; the remaining boundary tests are named for the general conversion/serialization behavior they own.

PR #515, opened 2026-07-31
Batch more database operations to improve the performance

MOZGIII:
Goes in after #513

---

This PR extends the work done in the couple previous PRs to get the performance of the execution to production levels. As of this PR, the benches demonstrate respectable 114% workflows processing rate versus the old DAG-based approach.

  | drain (25k) | throughput | vs DAG
-- | -- | -- | --
Pre-batching | ~26.8s | ~935/s | 0.36x (2.8x slower)
Legacy DAG | ~9.7s | ~2,586/s | 1.0x
Batchers / median | ~10.1s | ~2,475/s | ~0.96x
Batchers / best runs | ~8.5s | ~2,940/s | ~1.14x

<br class="Apple-interchange-newline">


piercefreeman (approved):

crates/lib/backend-postgres/src/workflow_completion/mod.rs
```diff
@@ -31,86 +31,91 @@ impl waymark_workflow_completion_backend::HasVmId for PostgresBackend {
...
+            }
+        }
+
+        let rows = sqlx::query(
             r#"
-            INSERT INTO vm_execution_results (vm_id, result)
-            VALUES ($1, $2)
+            INSERT INTO vm_execution_results (vm_id, result, error)
+            SELECT * FROM UNNEST($1::uuid[], $2::bytea[], $3::bytea[])
             ON CONFLICT (vm_id) DO UPDATE
```
piercefreeman, line 81: Could we deduplicate vm_ids before this upsert, and define the same-key behavior for the other positional batches? PostgreSQL rejects duplicate keys within one statement—for example, INSERT ... VALUES (1, ...), (1, ...) ON CONFLICT (vm_id) DO UPDATE ... fails with “ON CONFLICT DO UPDATE command cannot affect row a second time”—and the outcome batcher currently retries that permanent error forever.

crates/lib/execution-bringup/src/lib.rs
```diff
@@ -54,6 +54,30 @@ pub struct Config<NodeId> {
     /// Longest a snapshot waits to be batched before its write is flushed.
     pub snapshot_batch_delay: NonZeroDuration,
 
+    /// Maximum number of action-call requests coalesced into one batched
+    /// insert.
+    pub action_effect_reconciler_request_batch_max: NonZeroUsize,
```
piercefreeman, line 59: Could Config store one waymark_batcher::Policy per batcher instead of parallel *_max and *_delay fields? Callers reconstruct Policy immediately, while storing it directly removes the pairing boilerplate and miswiring risk.

MOZGIII: Good point! I'd adjust them all in a single corrective commit in a subsequent PR...

crates/lib/backend-postgres/src/workflow_completion/mod.rs
MOZGIII: This is a trade-off between doing the extra work and having sane edge-case behaviour.

I'll validate this again, but afaik I checked all the batched writes for logical impossibility of having duplicates; that concern is a global property that should allow us to skip the *local* correctness *enforcement*; the rationale is this is a hot path, and we should avoid doing extra work here; however - deduplication should not be prohibitive expensive here...

---

We already have a `read_batcher` that does one form of deduplication - data loader style; we could easily add a deduplicating writer with the similar model - so there are no technical difficulties to implementing it with the current batcher abstraction.

piercefreeman: Yeah, I like the semantics of keeping them separate (and of having the write one _not_ deduplicate by default)... But just a bit worried about query composition in the future if we run something that doesn't allow duplicate keys and this retries indefinitely.

Maybe we could solve at build-time by testing each query against a postgres db, or just forcing rigorous test cases for callers of this write function?

MOZGIII: The things are currently wired up through such that it'd be impossible by the virtue of having a special type and lack of a `Clone` at the workflow completion value passing channel; this is a type-system level guarantee.

However, I'd worried about *changes* being introduced here, but I don't yet know how we'd solve this; having type-system level guardrails makes it impossible to construct a failing test case.

---

A proper solution would probably be to handle the failure and check for duplicates before we retry - and if we find them either error out loudly, or just drop the ones that are late - for the workflow completions we have first-wins semantics.

MOZGIII: I've implemented a `deduplicating_write_batcher` for this case and other batch writes where this could've been an issue.

PR #522, opened 2026-08-04
Turn some interpreter errors into catchable exceptions

MOZGIII:
Goes in after #521.

---

This PR makes certain error conditions catchable via exceptions rather than exposing them via interpreter errors. The design removes the error conditions that are no longer present as errors from the interpreter error enums, and unifies the exception type-id definitions.

We also correct the overly-permissive behavior of the benchmarks/smoke/internal tests that allowed errors where they should've actually been not allowed (due to compile errors and VM runtime errors as opposed to caught exceptions).

---

We'd want to continue that work by sharing the exception types with compiler and maybe providing more exception details (i.e. exception-type-ids -> exception-types, and making the exception details compileable-into-bytecode-statics in addition to being runtime-constructable); but that's for later

### DIFF
--- a/crates/lib/vm-exception-type-ids/src/lib.rs
+++ b/crates/lib/vm-exception-type-ids/src/lib.rs
@@ -1,0 +1,34 @@
+//! The exception type identifiers for the built-in runtime exceptions.
+//!
+//! The single source of truth for the type id spellings shared by
+//! the producers (the compiler lowerings and the interpreters) and
+//! the exception handler matching.
+
+#![warn(missing_docs)]
+
+/// The runtime exception type identifier for a division by zero.
+pub const ZERO_DIVISION_ERROR: &str = "ZeroDivisionError";
+
+/// The runtime exception type identifier for an operation applied to values
+/// of unsupported types.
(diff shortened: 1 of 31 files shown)


crates/lib/vm-exception-type-ids/src/lib.rs
```diff
@@ -0,0 +1,34 @@
+//! The exception type identifiers for the built-in runtime exceptions.
```
piercefreeman, line 1: Could these constants live in waymark-vm-runtime-exception instead of a new crate? That crate already owns the exception types and is already on the relevant dependency path, so this would keep one exception concept and remove the new manifest/workspace/lock wiring.

MOZGIII: This was the initial place for them, but they had to move because we are also emitting exceptions at compile-time, so we place them in a place that is not runtime-specific; also see the updated PR message - I've added a section on the future work in this domain so it's clearer why this  I lean towards this approach.

piercefreeman (approved):

PR #527, opened 2026-08-06
Clean up legacy runloop: phase 3

MOZGIII:
Goes in after #526

---

This PR leaves the docs untouched: we'd want to do a proper pass on them later on either way, so we don't bother with them now.

More advanced cleanup/rework of the web app will be done later when we get to rewrite it with on a new tech stack.


piercefreeman (approved):

piercefreeman:
Good reminder we have some work to do to improve the webapp for the new functionality.

PR #535, opened 2026-08-06
Switch to a more robust CaptureRuntimeView design

MOZGIII:
Goes in after #534

---

This is a slight change required for a proper programmatic composition of interpreters.

It also, resolves an architectural oddity that sit back in my mind for a while - since after this part of the design was introduced. Not the placement of the trait is proper and its composition with the `vm-runtime-core`'s `FullRuntimeView` type does not have abstraction spill in any direction.


MOZGIII:
Heads up: I've readjusted this design a bit to simplify the lifetimes; the trait now implemented for the view itself as opposed to the interpreter - this makes reborrowing implementations less odd.
In general, the design is the same as before, but more concise and eliminates one axis of freedom that wasn't really critical - while the implementation is no longer on the interpreter, it still has control over how the view is captured since it specifies the view type it wants to use. Given that it's actually hard to imagine how we'd reuse the same type with different capture machinery - this is for the best, as not the capture is naturally reused for `FullRuntimeView` and `()` - so less code to write.

piercefreeman (approved):

PR #556, opened 2026-08-20
Deliver the workflow outcome before waiting for VM to terminate at transient execution

MOZGIII:
This PR changes the ordering in which we orchestrate the workflow outcome and VM driver termination during transient execution - it now semantically matching what's supposed to happen in production (more or less).

The underlying resource leak is still present, but it's not an issue for the transient execution specifically (this case) because the cancellation of the tasks at rust-side will take care of that; in production though, the resources will leak, and we need a proper cancellation routines support to do the resource cleanup.

### DIFF
--- a/crates/lib/transient-execution-worker-stream-bringup/src/lib.rs
+++ b/crates/lib/transient-execution-worker-stream-bringup/src/lib.rs
@@ -135,6 +135,8 @@ pub fn execute(runtime: waymark_system_vm::Runtime, skip_sleep: bool) -> Execute
             ActionCallCorrelation,
         >::new(action_result_rx);
 
+    let cancellation = tokio_util::sync::CancellationToken::new();
+
     let waymark_transient_execution_bringup::Execution {
         workflow_outcome_rx,
         driver_handle,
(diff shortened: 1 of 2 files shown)


crates/lib/transient-execution-worker-stream-bringup/src/lib.rs
```diff
@@ -165,6 +164,11 @@ pub fn execute(runtime: waymark_system_vm::Runtime, skip_sleep: bool) -> Execute
         };
 
         let _ = out_tx.send(response).await;
+
+        cancellation.cancel();
+
+        let Err(err) = driver_handle.await;
+        tracing::warn!(?err, "vm driver exited");
```
piercefreeman, line 169: Is it possible to add a test to catch this regression for the future?

MOZGIII: Sure, I'll add it

MOZGIII: Here: https://github.com/piercefreeman/waymark/pull/556/changes/cbced521d523ec030bb7fcd99607488ffaf49579

Pretty much the repro code you've came up with.

piercefreeman (approved):

PR #557, opened 2026-08-21
Drop the WorkflowNodeResult and return the completion value as-is

MOZGIII:
The WorkflowNodeResult was a leftover of the DAG-originated workflow completion shape; VM doesn't have this restriction - and this legacy-compat logic just interferes with users trying to return their data in unexpected ways. So, we just drop it.


piercefreeman (approved):
Nice - this is great

PR #558, opened 2026-08-22
Classify terminal VM driver exits correctly

piercefreeman:
Completed workflows currently emit an ERROR because the generic driver-thread wrapper logs Waymark’s expected terminal sentinel before the execution layer classifies it. This makes successful runs look unhealthy and obscures real driver failures.

- Downgrade generic driver-return logging to debug while preserving the returned error.
- Keep `NoReadyFramesOrWaitingPromises` as the expected parked-completion path.
- Emit an error only for genuinely unexpected persistent driver exits.

### DIFF
--- a/crates/lib/execution-driver/src/lib.rs
+++ b/crates/lib/execution-driver/src/lib.rs
@@ -232,7 +232,11 @@ async fn drive_one<
             tracing::debug!("VM evicted, parking workload");
             pinned.unpin(waymark_workload_pinning_core::UnpinMode::Park);
         }
-        Evicted::DriverError(_) | Evicted::HandledElsewhere => {
+        Evicted::DriverError(_) => {
+            tracing::error!("VM driver failed");
+            drop(pinned);
+        }
+        Evicted::HandledElsewhere => {
             tracing::debug!("VM evicted, unpinning");
             drop(pinned);
         }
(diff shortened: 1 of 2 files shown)


crates/lib/execution-driver/src/lib.rs
```diff
@@ -232,7 +232,11 @@ async fn drive_one<
             tracing::debug!("VM evicted, parking workload");
             pinned.unpin(waymark_workload_pinning_core::UnpinMode::Park);
         }
-        Evicted::DriverError(_) | Evicted::HandledElsewhere => {
+        Evicted::DriverError(_) => {
+            tracing::error!("VM driver failed");
+            drop(pinned);
+        }
+        Evicted::HandledElsewhere => {
```
MOZGIII, line 239: I'd argue at this level we don't care why VM was evicted - it's only important that it is...

crates/lib/vm-driver-thread/src/lib.rs
```diff
@@ -173,7 +173,7 @@ where
                     Error::Thread(error)
                 }
                 Ok(Err(error)) => {
-                    tracing::error!(?error, "vm driver terminated");
+                    tracing::debug!(?error, "vm driver terminated");
```
MOZGIII, line 176: In think in here we might want to switch them both to debug - I planned to do a UX overhaul before we release, but didn't get to it - the plan us to inform the users of the errors / panics where we have access to their concrete shapes - such that we can avoid logging idle VM exits as user-facing errors for instance

PR #570, opened 2026-08-22
Adjust and cleanup the conversion API surface

MOZGIII:
Goes in after #569


crates/lib/worker-inline-compat/src/lib.rs
```diff
@@ -2,84 +2,86 @@
...
+/// carries.
+///
+/// # Panics
+///
+/// Panics when the outcome holds a pending promise: an in-process body
+/// that hands back a pending value names a promise no one can settle —
+/// a bug in the action body, not an outcome it produced.
+fn encode_result(outcome: waymark_action_runtime_core::ActionCallOutcome<ReadyValue>) -> Vec<u8> {
+    waymark_action_runtime_convert::Converter::try_convert(outcome)
+        .expect("an action body's outcome holds no pending promise")
```
piercefreeman, line 58: [P2] Report conversion failure instead of losing the accepted action

This `expect` runs inside the inline pool's detached `tokio::spawn`, whose join handle is discarded. If encoding fails, the task exits before sending a result, while the pool retains its channel sender. The accepted action therefore produces no completion, loss report, or pool error and remains unresolved. The new argument-decoding panics have the same ownership problem.

Reproduced on the combined stack by returning `Ok(ReadyValue::List(vec![Value::Pending(PromiseStateId(7))]))` from an inline action: queueing succeeds, the adapter panics, and a one-second timeout around `poll_complete()` expires. Returning an ordinary integer completes successfully. Returning a pending promise is an action-body bug, as documented, but the previous adapter reported that conversion failure instead of silently losing the action.

MOZGIII: Yes. The right thing to do here is to actually propagate the panics; I've added it.

piercefreeman (approved):

PR #571, opened 2026-08-22
Explicitly propagate action loss though the action completions reconciler

MOZGIII:
Goes in after #553


crates/lib/backend-postgres-migrations/migrations/0024_rename_action_call_completions_outcome.sql
```diff
@@ -0,0 +1,4 @@
+-- The completion blob stores the call's execution result (Ok = the
+-- outcome the action produced, Err = the execution failed to produce
+-- one), so the column is named for what it holds.
+ALTER TABLE action_call_completions RENAME COLUMN outcome TO execution_result;
```
piercefreeman, line 4: [P1] Preserve existing completion blobs when upgrading

This rename preserves the stored bytes, but the reader now expects `Result<ActionCallOutcome<ReadyValue>, ActionCallLossError>` where the previous writer serialized `ActionCallOutcome<ReadyValue>` directly. I reproduced the incompatibility with the actual `RmpCodec`: serializing `ActionCallOutcome::Value(ReadyValue::Int(42))` and reading it as the new type fails with ``unknown variant `Value`, expected `Ok` or `Err` ``. A newly written completion round-trips successfully.

When a revived workflow demands one of these existing completions, the durable poller returns the decoding error before delivery/acknowledgment, and its shutdown guard cancels the execution subsystem. Restarting leaves the incompatible row in place. If deployments must retain existing durable state, the serialized bytes need to be migrated as well. An explicit drain/reset policy would also resolve the rollout requirement; this is conditional on retaining old state. The reproduction was at the codec level, not a live Postgres upgrade.

MOZGIII: Do we care about the upgrade paths today? I didn't implement proper migrations, as I figured we don't have a need for them while we're in this pre-release stage...

piercefreeman: We're running the RC candidate in a prod env at the moment, so ideally we would have some basic migration path up. If very hard to do I can choose a window and then just restart all the long running jobs from scratch but a migration path is preferable.

MOZGIII: I'd rather skip a migration path for now - the reason is for this kind of thing we'd need to alter sqlx migration routine and build a more advanced one - one that allows rewriting rows from rust (ideally) while mid-migration. It's not hard per se, but I'd rather not pivot to this from the other work we do currently.

The problem is if we need it - we'd need to do it with high priority since it'll block testing.

We'd need such a system anyway - it's just that for now I've decided to skip it in order to expedite the release... This means we'll, effectively, launch with an upgrade path that would assume data wipe (for later upgrades) for some changes that our current migration system can't handle, and one such change will likely be the multi-language support; I think it'd fine for the early-stage release phase we're shipping though

piercefreeman: > This means we'll, effectively, launch with an upgrade path that would assume data wipe (for later upgrades) for some changes that our current migration system can't handle, and one such change will likely be the multi-language support

Once we cut the initial release, I would be hesitant to ship implicit data deletion that happens anytime you just update a package version (and sqlx upgrades automatically). I'm fine with this approach pre-release but after the public one we'll need this pathway supported going forward.

MOZGIII: I agree.

To clarify: "assume data wipe" would mean an explicit deployment step, and we'll always have to take care to either have a proper automatic migration or a guardrail that would refuse to boot if the new version may corrupt old database data - the latter case being that "an upgrade path that would assume data wipe".

The principal issue is whether we promise a guarantee a proper upgrade path for *all* version upgrades, or we reserve the option to tell users "hey, you'll have to start with a fresh state now, and we can't migrate your data safely" for *some* version upgrades. I think the latter is just unavoidable for an early-days project like waymark, and also not too crazy expectation-wise.

piercefreeman (approved):

PR #574, opened 2026-08-22
Generalize the action-runtime-convert around the actual value type

MOZGIII:
Goes in after #573

---

Note that somewhere between this and ea167f3f1 we have become ~30% slower; to do: the bisect on benches later.


piercefreeman:
Is it worth adding some benchmark monitoring in CI to avoid this regression?

MOZGIII:
Yes, but benchmarking at CI can be really unstable - so it'll have to be properly build out; I'd suggest also adding a chaos testing environment and CI checks on each commit - i.e. when we kill postgres mid-execution and assert progress is not lost, etc. These two are related in a way that the work to do is similar, as it'll probably require a separate, more stable environment than Github Actions.

Oh, and executable size / compile time monitoring would be nice - so we can track now both drift over time and decide on when to address it.

For this current regression though - I'll just do a bench git bisect later when I get to fixing it, and address precisely where we get this slowdown. There is no opportunity loss here since we keep benchmarks operational throughout the dev work, so it can be relied on later.

piercefreeman (approved):

PR #605, opened 2026-08-25
Generalize spread fan-out to accept workflow helper functions

### LINKED ISSUE #577: Spread gather: generalize `SpreadExpr` to accept workflow helper functions
`await asyncio.gather(*[self.helper(i) for i in items], return_exceptions=True)` fails at build time: "Spread pattern element must be an @action call" (`python/src/waymark/ir_builder.py:3311`).

This is structural: `SpreadExpr.action` in `proto/ast.proto` is an `ActionCall`, and the Rust conversion requires it (`crates/lib/vm-ast-old-proto/src/lib.rs:243-251`). `ParallelExpr` already carries `Call` (action|function), so the fix is to generalize `SpreadExpr` the same way and teach the spread lowering to start function calls.

Direct-await helper laddering (`await self.helper(n)`, helpers calling helpers) already works — this closes the gap for fan-out.

MOZGIII:
Goes in after #574

---

Closes #577

This is a very trivial diff - changes are only at the compiler side, and in essence it was just a missing feature - must've been a limitation of the old DAG design, or just slipped through the cracks when the original IR was built...


MOZGIII:
Backported and merged to `main` ahead-of-stack

PR #607, opened 2026-08-26
Fix stale python proto codegen

MOZGIII:
This is a small fix for the codegen. It's quite odd, and we should be catching these kinds of desyncs at CI. This one, however, is harmless - just an out-of-sync docstring...


piercefreeman:
I don't think we're checking for this specific criteria at CI; only implicitly checking in unit tests if we happen to cross the protobuf path. I can add a CI step for sanity checking current generated code if we're seeing more drift than we expect.

piercefreeman (approved):

PR #609, opened 2026-08-27
Migrate well-known proto types to prost-wkt-types

MOZGIII:
Goes in after #607

---

This PR migrates from `prost-types` to `prost-wkt-types` - the best solution to the issue of lacking capabilities of the `prost-types` I know; this allows us to use proper well-known google types and resolved a ton of headaches with type conversions (proper `Any` support, proper `Timestamp` conversions, etc). All this without bringing our own unique types - so other crates can implement traits for these types - and they often do where it's relevant.

This is required for the use of proper `Timestamp` messages at the upcoming new scheduler proto.


piercefreeman (approved):
TIL

PR #620, opened 2026-08-31
Add finally support to Waymark IR

piercefreeman:
Adds finally support to Waymark IR and lowers each try/except/finally block into one VM exception scope.

The VM registration now carries an optional finalizer state. Jumps, returns, and raises all pass through the same unwind stack, so the finalizer bytecode is emitted once and resumed with the pending transfer.

Includes compiler snapshots and runtime coverage for nested finalizers, handlers, loop control, returns, and exceptions.


MOZGIII:
Noticed this and took a quick look at the bytecode - seems like this just compiles finally branch a bunch of times - would make more sense if we prepared a finally state and jump into it... Unless there is a reason not to - which there could be on a deeper look, exceptions are tricky.

PR #625, opened 2026-09-01
Add lossy-batcher

MOZGIII:
Goes in after #624

---

This is a new primitive - in addition to the already-existing `batcher` - that we'll use for the observability data batching; the nature of the observability data is often such that it's never allowed to slow down the main execution routines, and it's not critical if we loose some of it (it's not ideal either - so we don't aim to actively loose it, or can "sacrifice" it - but it's not as bad as losing operational data).


crates/lib/lossy-batcher/src/lib.rs
```diff
@@ -0,0 +1,341 @@
...
+    }
+
+    // Close the intake: every later push is refused and counted `closed`;
+    // a final non-empty filling buffer goes out best effort (counted
+    // `full` when no buffer is free, like any other swap).
+    if let Some(outcome) = shared.swapchain.close() {
+        shared.record_swap(outcome);
+    }
+    // Let the flushers drain what is buffered, then stop.
+    full_rx.lock().await.close();
```
piercefreeman, line 294: This can deadlock: an idle flusher holds `full_rx` across `recv().await`, but after shutdown wins the select that future is no longer polled, so this lock never becomes available. Repro: spawn with one flusher, yield until idle, then signal shutdown; awaiting the task times out even with zero events.

PR #630, opened 2026-09-05
Add observability-store-postgres (and -migrations)

MOZGIII:
Goes in after #629


crates/lib/observability-store-postgres/src/essential_metrics/query.rs
```diff
@@ -0,0 +1,177 @@
...
+    type NodeId = waymark_ids::NodeId;
+}
+
+impl waymark_essential_metrics_query_backend::Latest for Store {
+    type Error = sqlx::Error;
+
+    async fn latest(&self) -> Result<Vec<NodeSample<waymark_ids::NodeId>>, sqlx::Error> {
+        let rows = sqlx::query(&format!(
+            r#"
+            SELECT DISTINCT ON (node_id) {NODE_SAMPLE_COLUMNS}
```
piercefreeman, line 117: Scaling concern: fetching the latest sample per node scans retained history on every poll. Repro: seed 10,000 samples for each of 20 nodes and run `EXPLAIN (ANALYZE, BUFFERS)` on this query; my plan scanned/sorted all 200,000 rows to return 20. Cost grows with retention, not just node count.

PR #635, opened 2026-09-05
Port the soak harness to essential metrics

MOZGIII:
Goes in after #634


crates/bin/soak-harness/src/main.rs
```diff
@@ -57,12 +61,23 @@ async fn main() -> Result<(), color_eyre::eyre::Report> {
         .await
         .wrap_err("run migrations before soak")?;
 
+    let observability_pool =
+        waymark_sqlx_postgres_schema_pool::connect(args.dsn.expose_secret(), OBSERVABILITY_SCHEMA)
```
piercefreeman, line 65: This ignores `WAYMARK_OBSERVABILITY_DATABASE_URL`, while the child worker inherits and honors it. Repro: use main DSN A and observability URL B. The worker writes samples to B, but the harness resets/reads A, so readiness monitoring times out or consumes stale samples.

PR #648, opened 2026-09-05
Implement observability events backends at observability-store-postgres

MOZGIII:
Goes in after #647


crates/lib/observability-store-postgres/src/observability_events/query.rs
```diff
@@ -0,0 +1,218 @@
...
+        let mut query = sqlx::QueryBuilder::new(format!(
+            r#"
+            SELECT {EVENT_COLUMNS}
+            FROM observability_events
+            WHERE node_id = "#,
+        ));
+        query.push_bind(params.node_id);
+        // Keyset: strictly past the position in stream order.
+        if let Some(after) = params.after {
+            query.push(" AND node_sequence > ");
```
piercefreeman, line 192: Producer sequence is not a commit watermark. Repro: commit sequence 1, read tail and retain cursor 1, then commit sequence 0 from a slower flusher. Every subsequent `node_sequence > 1` read permanently skips the stored event 0. Multiple flushers permit this commit order.

PR #652, opened 2026-09-07
Feed the observability events from the VM driver hooks

MOZGIII:
Goes in after #651


crates/bin/start-workers/src/main.rs
```diff
@@ -195,6 +195,7 @@ async fn main() -> Result<(), waymark_fn_main_common::Error> {
         bringup_config,
         Arc::new(backend.clone()),
         remote_pool,
+        Some(Arc::new(observability_events_emitter)),
```
piercefreeman, line 198: The event sink and execution subsystem share a shutdown token, so intake can close before VM stop hooks run. Repro: cancel with an active VM, let the sink process cancellation first, then let the driver exit; its `vm_stopped` event is dropped as closed. Observability needs to remain open until producers finish.

PR #653, opened 2026-09-07
Route the VM driver run logging through the hooks

MOZGIII:
Goes in after #652


crates/lib/vm-driver-hooks-tracing/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+waymark-vm-driver = { workspace = true }
+waymark-vm-driver-core = { workspace = true }
+waymark-vm-driver-hooks = { workspace = true }
+waymark-vm-runtime-effect = { workspace = true }
+waymark-vm-runtime-promise-core = { workspace = true }
+
+tracing = { workspace = true }
+
+[dev-dependencies]
+tracing-subscriber = { workspace = true }
```
piercefreeman, line 17: This direct dev-dependency is outside `deny.toml`'s allowed wrappers. Repro: run `make lint`; cargo-deny fails with `crate 'tracing-subscriber = 0.3.22' is explicitly banned`. The current rust-lint CI job fails the same way.

PR #654, opened 2026-09-08
Per-VM timeline of the observability events

MOZGIII:
Goes in after #653


crates/lib/observability-store-postgres/src/observability_events/query.rs
```diff
@@ -214,5 +269,59 @@ impl waymark_observability_events_query_backend::Tail for Store {
...
+        let mut query = sqlx::QueryBuilder::new(format!(
+            r#"
+            SELECT {EVENT_COLUMNS}
+            FROM observability_events
+            WHERE {VM_ID_PRESENT} AND {VM_ID_EXPRESSION} = "#,
+        ));
+        query.push_bind(params.vm_id);
+        // Keyset: strictly past the position in timeline order.
+        if let Some(after) = params.after {
+            query.push(" AND (at, node_id, node_sequence) > (");
```
piercefreeman, line 293: When used to follow an active VM, this cursor can permanently miss late inserts. Repro: read through an event at t2, then commit an earlier event at t1 from a delayed batch; resuming with the returned cursor never yields t1. Live consumers need backfill semantics, or an explicit settled-data-only contract.

PR #655, opened 2026-09-08
Observability state API and the instance view

MOZGIII:
Goes in after #654


crates/lib/observability-store-postgres-migrations/migrations/0004_observability_vm_instances.sql
```diff
@@ -0,0 +1,152 @@
...
+        FROM inserted
+        WHERE payload ? 'vm_id'
+          AND kind LIKE 'vm_driver.%'
+    )
+    UPDATE observability_vm_instances AS instance
+    SET stopped_at = first_stop.at,
+        stop_kind = first_stop.kind
+    FROM (
+        SELECT DISTINCT ON (stop.vm_id) stop.vm_id, stop.at, stop.kind
+        FROM batch AS stop
```
piercefreeman, line 109: This only considers stops in the current batch. Repro: commit stop sequence 1 before start sequence 0 for the same VM/node, possible with two flushers. The stop initially has no run to update, and the later start never revisits it: `stopped_at` stays NULL although both events are stored.

PR #666, opened 2026-09-09
Bring the observability subsystem up under the benchmark on request

MOZGIII:
Goes in after #665


crates/bin/benchmark/src/observability.rs
```diff
@@ -0,0 +1,97 @@
...
+    node_id: waymark_ids::NodeId,
+    shutdown_token: tokio_util::sync::CancellationToken,
+) -> Result<Observability, color_eyre::eyre::Report> {
+    let sampling_handle = waymark_metrics_bringup::start(([127, 0, 0, 1], 0))
+        .wrap_err("install the metrics recorder")?;
+
+    let config = waymark_observability_config::ObservabilityConfig::from_env(&dsn.into())
+        .wrap_err("read the observability config")?;
+
+    let pool = waymark_sqlx_postgres_schema_pool::connect(dsn.expose_secret(), SCHEMA)
```
piercefreeman, line 48: This reset/count store uses the main DSN even when the config above selects another observability database. Repro: enable benchmark observability with DSN A and `WAYMARK_OBSERVABILITY_DATABASE_URL=B`; events go to B, but cleanup and the final event count operate on A.

PR #670, opened 2026-09-15
Build and embed a hello-world SPA in the HTTP server

piercefreeman:
Scaffolding for SPA that's compiled as part of the rust build pipeline.


.node-version
MOZGIII, line 1: I think this should live at the top-level of the repo, and be set per whole worktree - the idea is we'd rather have a common single node.js version for general un-further-qualified JS work (i.e. not including things that need specific or multi- versions support)

webapp/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "waymark-webapp",
```
MOZGIII: I propose using yarn/npm/pnpm workspaces - we can set it up later, but ultimately I imagine we'll have a family of JS packages that would benefit from the shared organization via workspaces the same way rust and python do

In that sense, it'd make sense to organize the tree per language (rust being primary, so keeping more of a root spot):
- `/package.json` (only workspace definition + "empty" root package)
- `/js/app/<package>` - for applications, i.e. everything that needs a toplevel build into a bundle of files to put on a web server or cli/server-side executable
- `/js/lib/<package>` - for libraries that are only included as part of other libs/apps or published as libraries - i.e. that need to keep their build/bundle outputs in order to be consumable by other TypeScript code
- `/.prettierignore` - global prettier ignore rules for
- `/node_modules` - global shared node modules among all workspace projects
- `/package-lock.json` - global shared lockfile

For the webapp specificlally:
- `/js/app/web/package.json` - webapp specific package
- `/js/app/web/vite.config.ts` - vite config
- `/js/app/web/tsconfig.json` - tsc config

crates/lib/http-api/src/lib.rs
```diff
@@ -46,7 +46,9 @@ pub fn router(mount_path: &str, routes: aide::axum::ApiRouter) -> axum::Router {
         // Nesting at the root is not a thing in axum; the router already is
         // at the root.
         "/" => router,
-        mount_path => axum::Router::new().nest(mount_path, router),
+        // Isolate the API's fallback, including at the mount point with a
+        // trailing slash, from the surrounding SPA's fallback.
+        mount_path => axum::Router::new().nest_service(mount_path, router),
```
MOZGIII: I think this needs to a test-case locally, and I also have some doubts on whether this is the right thing to do - as in, with respect to the composition from the outside...

(UDP: if we do the second part - no test case needed)

---

Ultimately, the API shouldn't mention or be aware of the webapp - they're supposed to be abstracted, so while we do need for the webapp's SPA we should approach it as a completely independent and self-container property of the API - or, rather, a whole router composition... As in - either the API has to establish, for it's own sake, that it always mounts ignoring the fallbacks - in which case the `/` case route is not valid - or we might move this isolation boundary outside of the `http-api` and into the composition layer.

To that end - it will likely cause the `mount_path` to be required to be specified twice - but that is something I anticipated would happed, and just hoped to work around - but this is, if fact, what the proper composition requirements dictate - so we can just give up some elegance for correctness of asbtractions - which is much more imporant imo.

If you agree I can prepare a PR to inject in front of this one - so the composition is cleaner.

webapp/vite.config.ts
```diff
@@ -0,0 +1,13 @@
+import { defineConfig } from "vite";
+import { viteSingleFile } from "vite-plugin-singlefile";
+
+export default defineConfig({
+  plugins: [viteSingleFile()],
```
MOZGIII: Cool and definitely will do for now - but ultimately I'd suggest doing proper bundle serving; we can add it later though.

crates/lib/http-webapp/build.rs
```diff
@@ -0,0 +1,47 @@
+use std::{env, fs, path::PathBuf, process::Command};
```
MOZGIII: <details><summary>sorry wrong diff</summary>
<p>

Consider using https://crates.io/crates/rust-embed

https://pyrossh.dev/repos/rust-embed/files/examples/axum-spa/main.rs/

</p>
</details>

crates/lib/http-webapp/src/lib.rs
```diff
@@ -0,0 +1,33 @@
+//! The compiled webapp, embedded in the server binary.
+
+use axum::{
+    Router,
+    http::{StatusCode, Uri, header},
+    response::{Html, IntoResponse, Response},
+    routing::get,
+};
+
+const INDEX: &str = include_str!(concat!(env!("OUT_DIR"), "/webapp/index.html"));
```
MOZGIII: Consider using [rust-embed](https://crates.io/crates/rust-embed)

https://pyrossh.dev/repos/rust-embed/files/examples/axum-spa/main.rs/

Maybe we can address https://github.com/piercefreeman/waymark/pull/670/changes#r4020649284 easily right away

.github/actions/setup-webapp/action.yml
```diff
@@ -0,0 +1,12 @@
+name: Set up webapp
```
MOZGIII: We might want to right away call it `setup-js` - and maybe make it generic for the whole `/js` workspace if we do https://github.com/piercefreeman/waymark/pull/670/changes#paneldiscussion_r4020449293

crates/lib/http-webapp/src/lib.rs
```diff
@@ -0,0 +1,33 @@
...
+
+/// Serve the SPA for browser routes, including direct visits to nested pages.
+/// Merge alongside the API and health routers before starting the HTTP server.
+pub fn router() -> Router {
+    Router::new().fallback_service(get(index))
```
MOZGIII: The `.fallback_service(get(...))` over just `.fallback(...)` is clever but non-trivial, maybe add a comment? 

```suggestion
    // Using`.fallback_service(get(...))` over just `.fallback(...)` here to only serve `GET` (and `HEAD`) fallback
    // requests, and let the rest 404.
    Router::new().fallback_service(get(index))
```

.github/workflows/ci.yml
```diff
@@ -230,6 +237,13 @@ jobs:
         with:
           ref: main
 
+      - uses: actions/setup-node@v4
+        if: hashFiles('webapp/package-lock.json') != ''
```
MOZGIII: We should probably add a todo to drop this `if:`

crates/lib/http-webapp/build.rs
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    let output = PathBuf::from(env::var_os("OUT_DIR").expect("output directory")).join("webapp");
+    let mut command = if cfg!(windows) {
+        let mut command = Command::new("cmd");
+        command.args(["/C", "npm"]);
+        command
+    } else {
+        Command::new("npm")
+    };
```
MOZGIII: Could be easier with https://docs.rs/xshell

crates/lib/http-webapp/build.rs
```diff
@@ -0,0 +1,47 @@
+use std::{env, fs, path::PathBuf, process::Command};
+
+fn main() {
```
MOZGIII: It might make sense to have the SPA build-invocation machinery implemented in a separate crate, and then reused here via the `build-dependencies`; that way we can implement tests for it

PR #700, opened 2026-09-17
Drop the lock renewal's shutdown token

MOZGIII:
Goes in after #699


crates/lib/execution-bringup/src/lib.rs
```diff
@@ -352,25 +352,10 @@ pub async fn start<Spawner, Backend, WorkerPoolRequests, WorkerPoolCompletions>(
...
-                        tracing::info!("action-call request lock renewal drained");
-                    }
-                    result
-                }
-            }
-        }
-    });
+    spawner.spawn(
+        "action effect reconciler lock renewal",
+        waymark_action_effect_reconciler::renewal::run(renewal_params),
```
piercefreeman, line 360: If the completion writer exits with a fatal error, its request rows remain in the database. This loop keeps renewing those locks indefinitely, so supervisor.drain() never finishes and other workers cannot retry the requests. Renewal still needs a failure shutdown path.

MOZGIII: Thanks! It was discovered by my local code reviews system too, I will get to fixing it in a bit.

